PR #80, opened 2026-04-27
Add CFL-reachability program analyses (aa, civf, pa, taint, vf)

hdz284:
Adds five Datalog programs that encode CFL-reachability program
analyses to `example/program_analysis/`, joining the existing
`cspa.dl` and `csda.dl` benchmarks already in this directory:

- `aa.dl`    — Alias analysis for C/C++
- `civf.dl`  — Context-insensitive value-flow analysis
- `pa.dl`    — Points-to analysis for Java
- `taint.dl` — Taint analysis
- `vf.dl`    — Value-flow analysis

Each program uses arity-3 `(src, dst, label_index)` tuples with
`int32` columns and `.input`/`.printsize` directives matching the
existing FlowLog example convention. Sources are adapted from the
souffle-experiments artifact.

Notes:
- The auxiliary relation traditionally named `A`/`Abar` is renamed to
  `MatchA`/`MatchAbar` to avoid collision with the lowercase EDB `a`
  under FlowLog's case-insensitive identifier resolution.
- Front-end compilation (parse, stratify, plan) is verified for all
  five programs via `flowlog-compiler`. End-to-end runs require
  per-label CSV facts produced from a CFL graph (see the
  souffle-experiments `transfer_facts.py` preprocessing).

### DIFF
--- a/example/program_analysis/aa.dl
+++ b/example/program_analysis/aa.dl
@@ -1,0 +1,59 @@
+// Alias analysis (CFL-reachability formulation)
+//
+// Indexed-label encoding: each edge is (src, dst, label_index). EDB relations
+// `a`/`ra` (assign and reverse-assign), `new`/`rnew` (allocation and reverse),
+// `load_i`/`rload_i` (load and reverse-load with label index), and
+// `store_i`/`rstore_i` (store and reverse-store with label index) are loaded
+// from per-label CSV files produced from a CFL graph (see
+// `tools/transfer_facts.py` style preprocessing).
+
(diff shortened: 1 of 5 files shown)


hdz284:
Closing — these benchmarks originate from the souffle-experiments artifact and are Soufflé programs. Their correct home is `flowlog-rs/vldb26-result-reproduction/program/souffle/` (PR flowlog-rs/vldb26-result-reproduction#2, reopened). A FlowLog port can be revisited separately if wanted.

PR #84, opened 2026-05-03
tidy(profiler): inline single-use NodeProfile setters into build_node

hdz284:
## Summary

`NodeProfile` had 7 single-use setter methods — `update_id`, `update_name`, `update_block`, `update_fingerprint`, `add_tag`, `add_operators`, `add_parents` — each called *exactly once* from `NodeManager::build_node`. They added no encapsulation (fields are `pub(super)`, no validation, no invariants), and the imperative *"default + 7 setter calls"* pattern made the construction sequence harder to scan than a struct literal.

This PR:

- Drops all 7 setters from `NodeProfile`.
- Rewrites `build_node` to construct `NodeProfile` with a struct literal, pushing `fingerprint.map(|fp| format!("0x{:016x}", fp))` and `vec![tag.to_string()]` into the literal directly.
- Tightens `parent_ids: Vec<usize>` typing (was inferred via `collect`).
- Renames `if let Some(profiler) = profiler.as_mut()` → `if let Some(p) = ...` in `with_profiler{,_ref}` to avoid shadowing the outer binding.

Net `+14 / -53`.

## Why this is *safer*, not just shorter

With the setters, adding a new field to `NodeProfile` would require remembering to call a new `update_x` from `build_node` — silently buggy if forgotten. With the struct literal, omitting a field is a compile error.

## Validation

- `cargo check --workspace --tests` — clean
- Smoke benchmark (galen reasoning) within 1.5% of base
- All `NodeProfile` setters were single-use (`rg 'update_(id|name|block|fingerprint)\\|add_(tag|operators|parents)'` confirms only `build_node` was using them)
- No public API touched

## Provenance

Generated by an internal agentic readability loop (`minimalist`, trial-2, iter 2 of 3). Diff was hand-reviewed before opening this PR. Sister PR for parser readability changes from the same trial run is filed separately.

### DIFF
--- a/crates/flowlog-build/src/profiler/mod.rs
+++ b/crates/flowlog-build/src/profiler/mod.rs
@@ -42,8 +42,8 @@ pub(crate) fn with_profiler<F>(profiler: &mut Option<Profiler>, f: F)
 where
     F: FnOnce(&mut Profiler),
 {
-    if let Some(profiler) = profiler.as_mut() {
-        f(profiler);
+    if let Some(p) = profiler.as_mut() {
+        f(p);
     }
 }
 
@@ -52,8 +52,8 @@ pub fn with_profiler_ref<F, E>(profiler: &Option<Profiler>, f: F) -> Result<(), 
 where
     F: FnOnce(&Profiler) -> Result<(), E>,
 {
-    if let Some(profiler) = profiler.as_ref() {
-        f(profiler)
+    if let Some(p) = profiler.as_ref() {
+        f(p)
     } else {
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


hdz284:
Closing as redundant: superseded by #87, which makes the same `NodeProfile` setter → struct-literal change in `profiler/node.rs` from a later trial run on the consolidated `cleanups(...)` naming convention. The trivial `profiler::with_profiler{,_ref}` binding rename from this PR (s/profiler/p/) is the only delta and can be picked up in a follow-up if wanted — not worth keeping a duplicate review burden for.

PR #105, opened 2026-05-05
cleanups(errors): tidy diagnostic formatting + label boilerplate

hdz284:
Three small commits across `*/error.rs` files (planner, stratifier, typechecker). Each extracts a small named helper or adopts captured-identifier formatting; the diagnostic surface is byte-identical.

### Why

Diagnostic builders had repeated label-construction blocks across error variants. Each commit names the duplicated shape, leaving the per-variant code as a thin call.

### Commits

- **`typechecker/error.rs (+ common/config.rs)`** — group imports; switch `eprintln!("{}", x)` calls to captured-identifier `eprintln!("{x}")`.
- **`planner/error.rs (+ parser/logic/atom.rs)`** — hoist three duplicated diagnostic-builder blocks into a `user_diagnostic` helper; collapse `Atom::Debug` into a `Display` delegate (with a clarifying doc comment); idiomatic `filter/map/collect` for parsed arguments.
- **`stratifier/error.rs (+ parser/logic/comparison.rs)`** — extract a `primary_only` label helper; group `comparison.rs` imports with sister files; fix one missing semicolon.

### Validation

### Risk

Low. Diagnostic strings/labels are unchanged; only the construction code path moved.

---
_Authored by [groomer](https://github.com/azureuser/groomer); every commit individually judged `like` by the inline diff judge._

### DIFF
--- a/crates/flowlog-build/src/planner/error.rs
+++ b/crates/flowlog-build/src/planner/error.rs
@@ -81,40 +81,69 @@ impl Diagnostic for PlanError {
                 head_span,
                 rule_span,
                 var,
-            } => user_diagnostic(
-                self.to_string(),
-                (
-                    *head_span,
-                    format!("`{var}` is referenced here but never bound by a positive body atom"),
-                ),
-                (*rule_span, "in this rule".into()),
-                "every variable in the rule head must appear in a positive body \
-                 atom so its value is determined during evaluation",
-            ),
+            } => {
+                let mut labels = Vec::new();
+                if let Some(l) = primary_label(*head_span) {
+                    labels.push(l.with_message(format!(
+                        "`{var}` is referenced here but never bound by a positive body atom"
+                    )));
+                }
+                if let Some(l) = secondary_label(*rule_span) {
+                    labels.push(l.with_message("in this rule"));
+                }
+                CsDiagnostic::error()
+                    .with_message(self.to_string())
+                    .with_labels(labels)
+                    .with_notes(vec![
+                        "every variable in the rule head must appear in a positive body \
+                         atom so its value is determined during evaluation"
+                            .into(),
+                    ])
+            }
 
             PlanError::InconsistentAggregation {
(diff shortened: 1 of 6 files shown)


crates/flowlog-build/src/planner/error.rs
```diff
@@ -81,69 +81,40 @@ impl Diagnostic for PlanError {
...
+            } => user_diagnostic(
+                self.to_string(),
+                (
+                    *head_span,
+                    format!("`{var}` is referenced here but never bound by a positive body atom"),
+                ),
+                (*rule_span, "in this rule".into()),
+                "every variable in the rule head must appear in a positive body \
+                 atom so its value is determined during evaluation",
+            ),
```
HarukiMoriarty: Is it real worth to abstract this function? If so, do you think we should abstract for all error, not only in planner, but also in other crates.

crates/flowlog-build/src/stratifier/error.rs
```diff
@@ -22,6 +22,12 @@ fn rule_labels(rules: &[(usize, Span)]) -> Vec<Label<FileId>> {
         .collect()
 }
 
+/// Single-element label vec for diagnostics that only point at one span.
+/// Returns an empty vec for dummy spans rather than fabricating a location.
+fn primary_only(span: Span) -> Vec<Label<FileId>> {
+    primary_label(span).into_iter().collect()
+}
+
```
HarukiMoriarty: Is it real worth to do this abstraction?

HarukiMoriarty (changes requested):

hdz284:
Thanks for the review — both abstractions backed out in f167b3b. The same label-builder pattern lives in catalog/typechecker/parser/build error.rs files too, so going crate-wide is more change than this tidy PR justifies. Both files reset to `main-next`.

`make test` green.

HarukiMoriarty (approved):

PR #106, opened 2026-05-05
cleanups(planner): extract helpers across rule planner + transformations

hdz284:
Seven commits hoisting duplicated blocks behind named helpers in `crates/flowlog-build/src/planner/`. Net delete-heavy (−160 LOC across rule_planner + transformation modules).

### Why

The rule planner's prepare/sip/core phases share scaffolding (filter handling, branch duplication for lhs/rhs, premap fan-out) that had grown into separate near-identical inline blocks. Each commit picks one duplicated shape, names it, and adds a doc comment.

### Commits

- **`planner/transformation/flow.rs`** — extract `signatures_to_trans_args`; flatten `flow_over_exprs` with an early-return happy path.
- **`catalog/rule/modify.rs (+ planner/transformation.rs)`** — extract `validate_atom_rhs_indices`, `lookup_arg_vars`, `build_renamed_atom_copies`; add an `Arc<Collection>` builder; fix lying doc comments. *Drops `−1 .clone(), −2 .unwrap()`.*
- **`planner/rule_planner/prepare.rs (+ build/engine/batch.rs)`** — extract `apply_drop_one_arg` to fold three prepare-phase filter handlers into thin wrappers (largest single deletion in this PR).
- **`planner/rule_planner/sip.rs (+ catalog/rule/populate.rs)`** — flatten control flow with `let-else`; replace lhs/rhs branch duplication with a 2-element loop.
- **`planner/rule_planner/core.rs`** — extract a `labelled` trace pretty-printer; fuse duplicated premap branches via a for-loop; replace a fixed-point `loop` with `||` short-circuit (preserves sequential semantics).
- **`planner/rule_planner.rs (+ parser/logic/arithmetic.rs)`** — extract `rhs_atom_map` to dedupe rhs-atom map construction; drop a throwaway intermediate `Vec<Transformation>`.
- **`planner/arithmetic.rs (+ parser/logic/fn_call.rs)`** — switch fn-call arg parsing to `try_collect()`; tidy imports + inline format args.

### Validation

### Risk

Low–medium. The largest commit (`catalog/rule/modify.rs`) is ~230 LOC of code movement; reviewers may want to walk it carefully. Differential-dataflow operator chains and trace fingerprints are untouched. Each commit was inline-judged `like` (or `error`→keep when the LLM judge timed out — cheaper gates already cleared correctness on the same diff).

---
_Authored by [groomer](https://github.com/azureuser/groomer)._

### DIFF
--- a/crates/flowlog-build/src/planner/rule_planner/core.rs
+++ b/crates/flowlog-build/src/planner/rule_planner/core.rs
@@ -8,7 +8,9 @@
 use tracing::trace;
 
 use super::RulePlanner;
-use crate::catalog::{AtomArgumentSignature, AtomSignature, Catalog, JoinPredicates};
+use crate::catalog::{
+    ArithmeticPos, AtomArgumentSignature, AtomSignature, Catalog, JoinPredicates,
+};
 use crate::planner::PlanError;
 use crate::planner::{KeyValueLayout, TransformationInfo};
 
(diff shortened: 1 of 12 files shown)


crates/flowlog-build/src/planner/rule_planner/core.rs
```diff
@@ -40,26 +42,20 @@ impl RulePlanner {
...
-            if self.remove_unused_arguments(catalog)? {
-                trace!("Catalog:\n{}", catalog);
-                trace!("{}", "-".repeat(60));
-                continue;
-            }
-
-            // Fixed point reached - no more optimizations possible
-            break;
+            trace!("Catalog:\n{}", catalog);
+            trace!("{}", "-".repeat(60));
```
HarukiMoriarty: Keep original structure since I need trace log after apply semi join or remove unused arguments.

crates/flowlog-build/src/planner/transformation.rs
```diff
@@ -4,18 +4,30 @@
...
+/// Wraps `Collection::new` plus an `Arc` so call sites stay readable.
+fn arc_collection(fp: u64, name: &str, layout: &KeyValueLayout) -> Arc<Collection> {
+    Arc::new(Collection::new(
+        fp,
+        name.to_string(),
+        layout.key(),
+        layout.value(),
+    ))
+}
+
```
HarukiMoriarty: This is over engineered, I think no need have seperate function.

HarukiMoriarty (changes requested):

hdz284:
Thanks for the review — both addressed in 8b586da:

- `core.rs`: restored the original `if/continue` loop with per-pass `trace!` pairs.
- `transformation.rs`: inlined `arc_collection` back at all 8 callsites.

`make test` green. I'll rebase against `main-next` before merge.

HarukiMoriarty (commented):

PR #118, opened 2026-05-28
fix(codegen): bridge Spur/String at UDF boundary under --str-intern

hdz284:
## Summary

Under `--str-intern`, FlowLog string columns are passed around as interned `Spur` handles, but user UDFs are written in plain Rust and declare their string parameters/returns as `String`. The built-in operator lowering (`builtin_op_to_token`) already bridges this boundary correctly via `resolve(...)` on string args and `intern(&...)` on string-returning results. The **UDF lowering paths did not** — so any program combining `--str-intern` with a UDF that takes or returns `:string` failed to compile, with errors like:

```
error[E0308]: mismatched types
   --> src/main.rs:341:52
    |
    |   .push((data.clone(), time.clone(), 1_i32));
    |          ^^^^^^^^^^^^ expected `(i32, Spur)`, found `(i32, String)`
```

## Why this matters

The bug silently blocked a useful combination: string-heavy programs (the very ones interning is designed for) that also want to call a Rust helper on a string column. The existing `udf_string_intern` fixture was meant to cover exactly this case — but in commit 7a73dc3 (`feat: build in function`) its `compile_flags = --str-intern` was removed, demoting it to test only the non-intern path. That hid the bug rather than fixing it. Forcing `--str-intern` on across the full 105-fixture suite produced exactly 2 failures (`udf_string_intern`, `udf_types`), both hitting this same boundary.

## Fix

Apply the same `resolve`/`intern` wrapping that `builtin_op_to_token` already uses, in all four UDF emission sites:

* `fncall_to_token` — UDF used as a value-producing expression. Wrap each `:string` arg with `resolve((arg).clone()).to_string()` and the whole call with `intern(&...)` when the UDF returns `:string`.
* `build_kv_fn_call_predicate`, `build_join_fn_call_predicate`, `build_row_fn_call_predicate` — UDF used as a boolean predicate. Same arg wrapping; no return wrapping needed (returns `bool`).

Two new tiny helpers on `CodeGen` look up param/return types from the parsed `.extern fn` declarations (`udf_param_types`, `udf_return_type`). Feature markers `string_resolve` / `string_intern` are set so the runtime imports are emitted.

## Test fixture change

Re-enables `tests/fixtures/datalog-batch/udf_string_intern/compile_flags = --str-intern`. This was the fixture's original intent; the previous workaround is now obsolete.

## Verification

| Path | Result |
| --- | --- |
| `cargo build --release --workspace` | clean |
| `cargo test --release --workspace` | 215 unit tests pass |
| `tests/fixtures/run_compiler.sh` | **105 / 105** pass |
| `tests/fixtures/run_lib.sh` | **105 / 105** pass |
| Forced `--str-intern` across every fixture | **105 / 105** pass (was 103 / 105 before) |

No oracle / performance impact: the non-intern path emits identical code to before; the intern path now emits additional `resolve(...)` / `intern(...)` only where the UDF signature actually needs them.

## Scope

This PR is intentionally narrow — codegen + one test re-enable. Auto-enabling `--str-intern` based on EDB column types is a separate concern and would land in its own PR.

### DIFF
--- a/crates/flowlog-build/src/codegen/arg.rs
+++ b/crates/flowlog-build/src/codegen/arg.rs
@@ -462,11 +462,20 @@ impl CodeGen {
             .iter()
             .map(|fc| {
                 let fn_name = format_ident!("{}", fc.name());
+                let param_types = self.udf_param_types(fc.name());
+                if string_intern && param_types.iter().any(|t| *t == DataType::String) {
+                    self.features.mark_string_resolve();
+                }
                 let args: Vec<TokenStream> = fc
                     .args()
                     .iter()
-                    .map(|a| self.build_kv_args_arithmetic_expr(a, string_intern, None))
-                    .collect::<Result<_, _>>()?;
+                    .enumerate()
+                    .map(|(i, a)| {
+                        let token =
+                            self.build_kv_args_arithmetic_expr(a, string_intern, None)?;
+                        Ok(wrap_udf_arg(token, param_type_at(&param_types, i), string_intern))
+                    })
+                    .collect::<Result<_, CodegenError>>()?;
                 Ok(if fc.is_negated() {
                     quote! { !udf::#fn_name(#( #args ),*) }
                 } else {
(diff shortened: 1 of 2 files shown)


HarukiMoriarty (changes requested):
Looks good to me. The CI test is not passed due to cargo clippy error, fix it.

hdz284:
Fixed the `clippy::manual_contains` warnings (`param_types.iter().any(...)` → `param_types.contains(&DataType::String)`, 4 sites in `crates/flowlog-build/src/codegen/arg.rs`). CI is now green — `cargo clippy --all-targets -- -D warnings` and `cargo test` both pass. @HarukiMoriarty PTAL when you get a chance.

hdz284:
@HarukiMoriarty thanks for the catch — clippy fix pushed in 731e154 (replace `iter().any(|t| *t == DataType::String)` with `Vec::contains(&DataType::String)` at all 4 UDF emission sites in `crates/flowlog-build/src/codegen/arg.rs`). CI is green now (clippy + 215 unit tests + 105/105 e2e fixtures). Ready for re-review.

HarukiMoriarty (approved):

PR #126, opened 2026-06-04
feat(doop): Soufflé surface-syntax gap fills for DOOP integration

hdz284:
## Context

Bridging gaps so the DOOP framework's `flowlog-logic/` tree (~8.3 KLOC `.dl`, written in Soufflé dialect) compiles cleanly on flowlog. Probed top-down against `cpp -P`-preprocessed DOOP analysis (`facts/facts.dl + basic/basic.dl + analyses/context-insensitive/analysis.dl` → 3161 lines). DOOP runs the C preprocessor upstream via `bin/flowlog-mirror.py` + `CPreprocessor.groovy`, so flowlog only consumes already-preprocessed `.dl` — no preprocessor work in this PR.

This PR closes the **five** surface-syntax / inliner gaps that block the smoke probe. After this PR the smoke compiles through to line 3154/3161; the remaining failure is a deep semantic question about parent-scope name resolution for parameterized comps (deferred — flagged at the end).

## What's in this PR

| # | Gap | Touch points |
|---|-----|--------------|
| 1 | **Body-position aggregates** `r = op [expr] : { body }` | New `parser/logic/body_aggregate.rs` AST + grammar rules; new `parser/desugar.rs` pass that lifts each to a fresh `__body_agg_N` IDB relation with a normal head-position aggregate. Typechecker / subtype / populate keep `unreachable!` arms — desugar runs *before* them. Supports `count` K=0/K=1; K≥2 rejected with clean diagnostic. |
| 2 | **Parenthesized arithmetic at boundary** `(a - b) <= 15` | Grammar split `arithmetic_expr = { paren_arith \| unparen_arith }`; parens stripped at parse time so the AST stays flat. |
| 3 | **Negation w/ parens + bare boolean builtins** `!(P)`, `!match(...)`, `match(...)` as a body predicate | Grammar `negative_atom = "!" ~ ("(" inner ")" \| inner)`; bare/negated boolean builtins lowered at parse time to `Compare(call, =, true\|false)` so no new predicate variant ripples downstream. |
| 4 | **Soufflé regex builtin** `match(re, s) : bool` | New `BuiltinOperator::Match` + grammar `match_op` + codegen `flowlog_runtime::regex::Regex::new("^(?:re)\$")` (anchored, matches Soufflé semantics). `flowlog-runtime` re-exports `regex` so generated code works in both binary and lib modes without scaffolder plumbing. |
| 5 | **Inliner & dot-normalizer descend into body-aggregate inner bodies** | Previously `rewrite_rule` / `normalize_rule_dots` only walked `rule.rhs` top-level. Atoms inside `min v : { Atom(...) }` kept their pre-inlining names and the desugar pass couldn't find their declared type. Both passes now recurse via small `rewrite_predicate` / `normalize_predicate_dots` helpers. |

## DOOP-side fix (out of band, not in this PR)

`doop-flowlog/bin/flowlog-mirror.py::_PLAN_RE` was stripping only the first line of multi-line `.plan` directives, leaving orphan `2:(...), 3:(...)` continuations that broke parsing. Patched in the DOOP-flowlog repo.

## Tests

- `cargo clippy --all-targets -- -D warnings` clean
- `cargo test` → **193 passed**
- `bash tests/fixtures/run_compiler.sh` + `bash tests/fixtures/run_lib.sh` → **108 passed**, including two new fixtures:
  - **`comp_body_aggregate`** — regression for gap #5 (body aggregate inside a `.comp`).
  - **`_souffle_parity`** — **byte-identical output parity vs. Soufflé 2.5** on every construct in this PR. Expected/*.csv values were generated by Soufflé on the same Edge/Node inputs; flowlog must reproduce them exactly. This is the verification that the engine changes preserve Soufflé semantics, not just compile.

## What's deferred

- **Parent-scope resolution for parameterized comps** — when `ContextInsensitiveConfiguration` is instantiated as a type parameter of `BasicContextSensitivity`, its body's bare references should resolve in the enclosing `mainAnalysis` scope (where `isImmutableHContext` is declared). Implementing this cleanly would touch the inliner's scope chain non-trivially. This is the next concrete gap to bridge for full DOOP coverage.
- **Records / ADTs** (`.type Foo = [ ... ]`) — listed in STATUS.md as supported but not verified against DOOP yet.
- **K≥2 body-aggregate count** — DOOP never uses it; rejected with clear diagnostic.


hdz284:

PR #127, opened 2026-06-05
feat(doop): enclosing-scope resolution + expression atom args (round 2)

hdz284:
## Context

Stacked on #126. Continues bridging Soufflé surface-syntax gaps so DOOP's context-insensitive analysis compiles on FlowLog. Probed top-down against the `cpp -P`-preprocessed DOOP analysis, using **Soufflé 2.5 as the correctness oracle**.

> Base is `feat/doop-gap-fills` (#126), so the diff here is round-2 only. Retarget to `main-next` once #126 merges. (CI runs on PRs targeting `main`/`main-next`, so checks here are deferred until retarget — validated locally below.)

## What's in this PR

| # | Gap | Touch points |
|---|-----|--------------|
| 1 | **Enclosing-component relation resolution** — an unqualified relation ref inside a nested component that isn't a local `.decl` now resolves up the ancestor chain (closest wins; local shadows enclosing shadows global), matching Soufflé. Fixes `ContextInsensitiveConfiguration` (instantiated as `mainAnalysis.configuration`) referencing the parent `mainAnalysis`'s `isImmutableHContext` — the gap #126 explicitly deferred. | `parser/inliner.rs` (new `enclosing_relations` scope map) |
| 2 | **Expression-valued atom arguments** — `FormalParam(idx - 1, …)`, `ord(h)`, `as(x, T)` in body atom argument position. Grammar accepts arithmetic in `atom_arg`; a desugar pass lifts each to a fresh variable bound by an equality filter (mirrors `BodyAggregate` — `AtomArg::Expr` is transient, `unreachable!` after desugar). Expression args in **negated** atoms are rejected with a clear diagnostic. | `grammar.pest`, `logic/atom.rs`, `parser/desugar.rs`, `parser/program.rs` |
| 3 | **Cleanup** — module doc covers both passes; expr-arg pass in its own section; shared `segment_rules`/`segment_rules_mut` + `atom_has_expr_arg` helpers; allocation-free `predicates_have_expr_arg` fast-path (rules without expression args skip the rhs clone — zero-copy common case); dropped an obsolete import hack; `resolved_body_names` test helper. | `parser/desugar.rs` |

## Verification vs Soufflé 2.5

- **Enclosing-scope**: confirmed against Soufflé's `--show=transformed-datalog` — the child rule resolves `isImmutableHContext` to exactly `mainAnalysis.isImmutableHContext`.
- **Expression atom args**: new `expr_atom_args` fixture is **byte-identical to Soufflé 2.5**.

## Tests

- `cargo test` → **200 passed** (7 new: enclosing-scope shadowing matrix + expr-arg lift / plain-args-unchanged / negated-reject).
- `tests/fixtures/run_compiler.sh` → **109 passed** (binary) + lib mode.
- `cargo clippy --all-targets -- -D warnings` clean.

## Benchmark — no regression (these are compile-time-only passes; runtime IR is unchanged)

A/B via `flowlog-bench` `cross-flowlog-version`, base = `feat/doop-gap-fills` (#126) vs this branch:

| pair | base | head | time% | rss% | verdict |
|------|------|------|-------|------|---------|
| tc.dl=G20K-0.001 | 58.17s | 59.11s | +1.62 | +4.74 | OK |
| sg.dl=G10K-0.001 | 21.26s | 21.89s | +2.93 | +0.62 | OK |
| galen.dl=galen | 7.78s | 7.81s | +0.31 | −1.59 | OK |
| crdt.dl=crdt | 26.70s | 25.26s | −5.40 | +1.18 | OK |
| dyck.dl=postgre | 0.468s | 0.459s | −2.04 | −1.08 | OK |
| andersen.dl=medium | 0.677s | 0.625s | −7.70 | −5.40 | OK |

`dyck`/`andersen` are sub-second runs; their initial 3-run medians tripped the +10% gate purely from startup jitter (and contention with a concurrent build) — re-run at **15 samples** they land at −2.04% / −7.70%. **Output relation tuple counts are byte-identical base vs head on all six programs.**

## Deferred (next gap)

**Assignment binding** (`calleeCtx = callerCtx`, `heap = "…"`, `descriptor = cat(…)`) — FlowLog requires positive-atom range-restriction and rejects `=` as assignment. A rubber-duck critique showed this needs substitution **plus** zero-positive-rule → fact conversion (DOOP's pure-init rules become factual after substitution) and aggregate-result handling. Scoped for a focused follow-up with its own Soufflé-parity validation.


hdz284:

PR #128, opened 2026-06-05
feat(doop): declared-empty relations as empty EDBs (round 3)

hdz284:
## Context

Round 3 of the DOOP surface-compat gap fills, **stacked on #127** (base = `feat/doop-gap-fills-round2`; retarget to `main-next` once #126/#127 merge). Probed top-down against the `cpp -P`-preprocessed DOOP context-insensitive analysis, with **Soufflé 2.5 as the correctness oracle**.

After #127, two *driver-side* issues had to be cleared first to reach the next engine gap (documented in `doop-flowlog`, not engine bugs): a missing `-DCONFIGURATION=ContextInsensitiveConfiguration` cpp define, and the `--str-intern` flag (DOOP uses `ord`). The next hard engine error was:

```
error: rule 684 references relation `mainanalysis·heapallocation_keep`, which is not yet defined at this point in the program
   !HeapAllocation_Keep(heap).
```

## The gap

A relation that is declared (`.decl`) but has **no defining rule, no inline facts, and no `.input`** is, in Soufflé, simply an empty relation: a negated use `!R(x)` is always satisfied and a positive use `R(x)` never fires. DOOP relies on this — `HeapAllocation_Keep` in `string-constants.dl` is declared and used only as `!HeapAllocation_Keep(h)`.

FlowLog pruned such a relation out of `needed_preds` as an "underived IDB" but kept the rules referencing it, so the stratifier raised `ForwardReference`.

## Fix

`prune_dead_components` now registers every declared relation that is still referenced by a **needed** rule but has no rule/facts/`.input` as an **empty inline-fact EDB** (a zero-row `facts` entry). It is then available before the first stratum, codegen emits an empty collection, and negation/positive use behave exactly as in Soufflé. References reachable only through pruned (dead) rules do **not** resurrect the relation. Undeclared references are still rejected earlier by `validate_relation_references`.

Two codegen guards are tightened so a zero-row sentinel isn't mistaken for real inline data (both **behaviour-identical** for programs whose `facts` entries all hold real tuples):
- binary preload (`io/input.rs`): preload only when some fact vec is non-empty.
- relation codegen (`relation.rs`): import `SEMIRING_ONE` only when some EDB has a non-empty inline fact (matches `gen_inline_facts`, a no-op for empty) — avoids an unused-import error.

## Tests

- New **`empty_relation_negation`** fixture: `expected/*.csv` generated by **Soufflé 2.5**; byte-identical in **both binary and lib mode** (negated use == all input, positive use == empty).
- Two unit tests: declared-underived-but-referenced relation becomes an empty EDB; a dead-only reference does not revive it.
- `cargo test` → **202 passed**; `cargo clippy --all-targets -- -D warnings` clean; full fixture suite green.

## Benchmark (no regression)

`flowlog-bench` `cross-flowlog-version`, base = `main-next` (`20f2aa0`) vs this branch (`ee2613f`), 3 runs/pair, workers=32 — **every pair within tolerance** (time +10%, RSS +20%):

| pair | base | head | time% | rss% |
|------|------|------|-------|------|
| tc.dl=G5K | 1.66s | 1.73s | +4.21 | +0.59 |
| tc.dl=G10K | 9.22s | 9.17s | −0.58 | +6.24 |
| sg.dl=G5K | 3.10s | 2.92s | −5.80 | +0.09 |
| sg.dl=G10K | 20.55s | 20.89s | +1.62 | +0.68 |
| galen.dl=galen | 5.62s | 5.47s | −2.82 | +0.34 |
| crdt.dl=crdt | 2.92s | 2.92s | −0.21 | +2.42 |

These are compile-time-only source-to-source passes, so runtime IR is unchanged for programs that don't use the new behaviour; deltas are bidirectional noise.

## Next gap (deferred)

The smoke now reaches **assignment binding** (`t = "boolean"`, `calleeCtx = callerCtx`) — a variable defined by an equality with no positive-atom range restriction. A substitution-desugar draft exists; it additionally needs empty-body→fact conversion + multi-head-fact splitting. Scoped as a focused follow-up with its own Soufflé-parity validation.


hdz284:

PR #129, opened 2026-06-05
feat(doop): Soufflé surface-syntax compatibility — DOOP compiles + runs end-to-end

hdz284:
**Consolidates the former #126 → #127 → #128 → #129 stack into one PR** against `main-next` (four logical commits). Closes every surface-syntax / semantic gap that blocked DOOP's `cpp`-preprocessed **context-insensitive** analysis from compiling on flowlog. Soufflé 2.5 is the correctness oracle throughout.

## Result

The real preprocessed DOOP analysis now compiles **end-to-end** and runs:

```
flowlog-compiler --str-intern --output-dir out assembled.dl   # parse→ground→stratify→codegen→cargo build → EXIT 0, 168 MB binary
./assembled -w 4                                              # runs the dataflow, EXIT 0
out/  →  all 21 DOOP outputs: VarPointsTo, CallGraphEdge, Reachable, InstanceFieldPointsTo, StaticFieldPointsTo, …
```

(Outputs are empty in the probe only because it used an empty fact dir.)

## Commits (one per gap class)

1. **Soufflé surface-syntax gaps** — body-position aggregates, parenthesized arithmetic at boundaries, negation-with-parens + bare boolean builtins, `match(re,s)` regex. Body aggregates are **lowered to flowlog-native head-position aggregations** (`agg(e)` over an aux IDB) — surface-syntax acceptance, *not* a parallel aggregate runtime.
2. **Enclosing-scope relation resolution** (ancestor-chain visibility) + **expression-valued atom arguments** (`R(idx-1, x)`, `ord(h)`, `as(x,T)`).
3. **Declared-empty relations as empty EDBs** — Soufflé treats a declared rule-less relation as empty; `!HeapAllocation_Keep(h)` now resolves.
4. **Assignment binding (grounding)** — `isModifier(x) :- x = "abstract"` etc.; substitute equality-bound vars + relocate ground rules to facts. Plus the codegen fixes this uncovered (fact-liveness pruning, EDB element-type pinning, `unused_variables` lint exemption).

## Tests

- Per-feature Soufflé 2.5-parity fixtures: `comp_body_aggregate`, `_souffle_parity`, `expr_atom_args`, `empty_relation_negation`, `assignment_binding` (each byte-identical to Soufflé in binary + lib mode).
- `cargo test` → **204**; `cargo clippy --all-targets -- -D warnings` clean; **full fixture suite green: binary 111/111, lib 111/111**.

## Benchmark (no regression)

`flowlog-bench cross-flowlog-version`, base = `main-next` (`20f2aa0`) vs this branch, 3 runs/pair, workers=32 — every pair within tolerance (time +10%, RSS +20%). These are compile-time-only source-to-source passes; runtime IR is unchanged for programs that don't use the new syntax.

## Notes

- Remaining for a full Soufflé swap-in is **integration, not language gaps**: wire a `FlowLogAnalysis` DOOP backend (pass `-DCONFIGURATION=…` + `--str-intern`) and diff `VarPointsTo` vs Soufflé on a real fact corpus.
- Pre-existing, out of scope: a relation name that lowercases to a Rust keyword (e.g. `In` → `in`) breaks codegen; DOOP doesn't hit it.


hdz284:
Superseded by #130 (feat/doop-end-to-end): a single clean-commit consolidation that drops the Soufflé-style body-aggregate detour (DOOP uses no aggregates, so it is unneeded) and is now verified byte-identical to Soufflé 2.5 across all 26 DOOP outputs on 9 DaCapo apps. Consolidating to one PR.

PR #130, opened 2026-06-05
feat(doop): consolidate DOOP end-to-end gap-fills into one clean branch (supersedes #126–#129)

hdz284:
## What

A single clean commit that bridges every Datalog/Soufflé surface-syntax gap needed for
DOOP's context-insensitive pointer analysis to compile and run **end-to-end** through
FlowLog as a drop-in Soufflé replacement — while keeping **FlowLog-native head-level
aggregation** (no Soufflé-style body aggregates).

This **consolidates the merits of the stacked PRs #126 → #127 → #128 → #129 into one
branch** and drops the body-aggregate detour. Intended to **supersede** that stack.

## Merits merged (verified present in this branch)

| Source | Merit | Key files |
| --- | --- | --- |
| #126 | Parenthesized arithmetic (`Factor::Group`) | `parser/logic/arithmetic.rs`, `catalog/arithmetic.rs`, `planner/arithmetic.rs`, `codegen/arg.rs`, `typechecker/` |
| #126 | `match(re, s)` regex builtin (anchored, Soufflé semantics) | `parser/logic/builtin.rs`, `codegen/features.rs`, `compiler/scaffold.rs` |
| #126 | `!(Atom)` negation sugar + bare bool builtins | `parser/logic/predicate.rs`, `parser/logic/rule.rs` |
| #127 | Enclosing-scope / ancestor-chain relation resolution | `parser/declaration/comp.rs`, `parser/declaration/directive.rs` |
| #127 | Expression-valued atom args (`R(idx-1, x)`, `ord(h)`, `as(x,T)`) | `parser/desugar.rs`, `parser/inliner.rs`, `codegen/arg.rs`, `catalog/rule.rs` |
| #128 | Declared-but-underived relations → empty EDBs (Soufflé parity) | `parser/program.rs`, `stratifier/core.rs`, `codegen/edb_handles.rs` |
| #129 | Equality-as-assignment binding / grounding | `parser/desugar.rs`, `parser/program.rs` |

## Intentionally dropped

- **Soufflé-style body aggregates** (`r = op [e] : { body }`, `desugar_body_aggregates`,
  `Predicate::BodyAggregate`) introduced in #126. FlowLog uses its native head-position
  aggregates. Confirmed absent on this branch (`grep BodyAggregate` → none).

## Verification

- `cargo test --workspace --release` — **green**. `flowlog-build` runs **200** unit tests
  including `equality_assignment_alias_substituted_into_head`,
  `equality_assignment_const_only_becomes_fact`,
[...]

PR #131, opened 2026-06-06
fix(codegen): escape relation names that collide with Rust keywords

hdz284:
## What

A relation whose name is a Rust keyword (`type`, `match`, `in`, `loop`, …) crashed
codegen with an opaque `syn` panic (`valid token stream`), because the relation
name was emitted verbatim as a binding in generated code — `let type = …`, which
is not valid Rust. DOOP and general Datalog programs are free to name relations
this way, so this is a real robustness gap (it surfaced while hardening the
DOOP↔FlowLog integration).

Minimal repro (pre-fix → panic):

```
.decl Type(a: symbol, b: symbol)
.input Type(IO="file", filename="t.csv", delimiter=",")
.decl Out(a: symbol)
Out(a) :- Type(a, _).
.output Out
```

## Root cause

Two codegen sites turned a relation name into its collection-binding ident with
`format_ident!("{}", name)`, bypassing the canonical `rust_ident` sanitizer
(whose own doc says *"All call sites that embed a user name as a Rust ident must
route through this"*):

- `codegen/ident.rs` — `make_global_ident_map`
- `codegen/edb_handles.rs` — the EDB `(handle, collection)` binding

Separately, `rust_ident` itself had a **latent panic**: it called
`Ident::new_raw` for `crate`/`self`/`Self`/`super`, but those four keywords
*cannot* be raw identifiers (`r#self` is rejected) — so a relation/field named
that way (e.g. as an `.output`) would also crash.

## Fix

- `rust_ident`: escape the four non-raw-able keywords with a trailing underscore
  (`self` → `self_`); all other keywords still become raw idents (`type` →
  `r#type`).
- Route the global ident map and the EDB collection binding through `rust_ident`.
- `pub(crate)` the `relation` module so codegen can reach the helper.

**No behaviour change for non-keyword names** (the overwhelming common case):
`rust_ident` returns them unchanged, so generated code is byte-identical.

## Tests

- Unit tests (`build/relation`): every strict + reserved Rust keyword yields a
  binding that parses as valid Rust (`let <id> = 1; let _ = <id>;`), escapes are
  distinct, and non-keywords pass through. The `every_keyword_yields_usable_binding`
  test would have caught the old `new_raw` panic for the four special keywords.
- New `keyword_relation_names` datalog-batch fixture: joins relations named
  `Type` and `Match` end-to-end, byte-identical expected output.
- `cargo test --workspace --release` green (197 lib tests); `cargo clippy
  --all-targets -- -D warnings` clean; 15 diverse fixtures (joins, negation,
  aggregation, `.comp`, recursion, multi-head) green.
- Verified end-to-end: relations named `Type`, `Match`, `In`, `Super`, `Loop`,
  `Move` now compile **and** run (EDB, IDB, recursion, and `.output` positions).

## Interaction with #130

Independent of the DOOP surface-syntax work in #130 (different files; cherry-picks
cleanly with no conflicts). Verified the composed tree compiles + runs the DOOP
context-insensitive analysis on `batik` with **VarPointsTo = 37,237,353** (all 26
relations unchanged) — zero regression. Kept as its own PR to keep #130 focused;
can target `main-next` independently.


HarukiMoriarty:
The `syn`-panic fix is correct, but I think the underlying method still has a bug
and we'll keep hitting it one keyword at a time. Flagging the root cause + a
suggested fix.

### Root cause

We derive the internal Rust binding from the user's relation name by escaping it
*per-name*. That can't be made collision-free:

```rust
if matches!(name, "crate" | "self" | "Self" | "super") {
    format_ident!("{}_", name)   // crate -> crate_
}
```

`crate_` is an ordinary, legal relation name a user (or a synthesized relation)
can already have, so two distinct relations collapse to one binding:

| relation | keyword? | emitted ident |
|----------|----------|---------------|
| `crate`  | yes      | `crate_`      |
| `crate_` | no       | `crate_`      |

→ two `let crate_ = …`, one silently shadows the other: a **silent miscompile**,
worse than the panic we replaced. `r#crate_` doesn't help — for a non-keyword
`r#X == X` in Rust (`let r#crate_ = 1; println!("{}", crate_)` prints `1`). A
per-name escape fundamentally can't guarantee uniqueness because it can't see the
other names.

### Suggested fix

Stop deriving the code binding from the user name — the same thing Souffle does
for C++ keywords. Emit a synthetic ident and keep the human name only as a string
for I/O. We already have all the pieces on `Relation`: `raw_name` (user-facing,
for filenames) and a unique `fingerprint` / stable declaration order for `<N>`.

1. **Now (small, makes it sound):** add a uniqueness guard in
   `make_global_ident_map` that *fails loud* on a duplicate ident, so the silent
   miscompile can't ship.
2. **Follow-up (real fix):** bind relations as `rel_<N>_<sanitized name>` and drop
   the keyword-escape special-case entirely — collisions and keywords both become
   impossible by construction. Two constraints: assign `N` from a deterministic
   order (declaration order over `program.relations()`) to keep the byte-diff
   fixtures stable, and route *every* binding site through a single
   `binding_ident(rel)` helper (this PR already showed what one bypassing site
   costs). `rust_ident` then stays only for user-facing idents (`rel::Foo`
   aliases, `Inputs` fields), with a uniqueness assertion there too.

Happy to take the follow-up.

hdz284:
Good catch — agreed the per-name escape can collapse `crate` and `crate_` into one binding, which is a silent miscompile, not just a panic.

Implemented the "make it sound now" half in 09989bc: `make_global_ident_map` now routes through a `build_global_ident_map` helper that **fails loud** when two distinct relations (distinct fingerprints) lower to the same binding ident, naming both offenders. Since that fingerprint→ident map is the single source both the EDB collection bindings and `find_global_ident` read from, guarding it covers every collection-binding site.

Verified end-to-end: a program declaring `crate` and `crate_` now aborts codegen with

```
codegen: relations `crate` and `crate_` both lower to the Rust binding `crate_`; per-name keyword escaping cannot guarantee unique collection idents. Rename one relation. ...
```

instead of miscompiling. Added unit tests covering each non-raw-able keyword vs. its underscore twin, the repeated-relation (same fingerprint) non-collision case, and distinct keywords (`type`/`match`) coexisting.

Left the structural fix (synthetic `rel_<N>_<name>` binding idents that make collisions + keywords impossible by construction) as the follow-up — happy for you to take it as offered, or I can pick it up.

PR #135, opened 2026-06-08
perf: FlowLog string-handling optimizations (output resolve + direct writes + FxHash interner)

hdz284:
A holistic pass over FlowLog's string handling under `--str-intern`. Three changes, each correctness-neutral (output is byte-for-byte identical):

---

## 1. Interned-string `.output` — flat resolver (the big win)

Users reported Soufflé string `.output` is much faster than FlowLog, while `.printsize` is fast. The two differ only at the **sink**, so the gap is entirely drain-time.

**Root cause:** with `--str-intern`, string columns are `lasso::Spur`, and the drain resolved **every cell** through `ThreadedRodeo::resolve` — a `DashMap` lookup that **hashes the key (SipHash) and takes an `RwLock` read guard per call**. `.printsize` never resolves.

Micro-benchmark (56M resolves): `ThreadedRodeo::resolve` **11.7 M/s** vs flat `Vec<&str>` index **828 M/s** (**70×**).

**Fix:** `Spur` keys are dense in `[0,len)`, so resolve at drain time through a flat `Vec<&'static str>` snapshot (`resolve_out`) indexed by key — O(1), no hash, no lock. Built lazily after fixpoint (complete in batch mode; incremental epochs' newer keys bounds-check-fall-back to live `resolve`). Mid-dataflow `resolve` (joins/UDFs) is untouched. Gated by a `string_resolve_out` feature so `.printsize`-only programs pay nothing.

## 2. Direct row writes

File rows write string columns' UTF-8 bytes via `write_all` (skipping `core::fmt`/`writeln!`); numeric/bool columns still format through `write!`. Output `BufWriter` enlarged to 1 MiB to amortize syscalls.

## 3. FxHash interner (free micro-opt)

`ThreadedRodeo` defaulted to `RandomState` (SipHash); the interner's keys are program-controlled, so HashDoS resistance is wasted while per-byte cost is paid on every `intern`/`resolve`. Switched to `rustc_hash::FxBuildHasher`. Isolated interning (real batik strings): **3.6 → 5.6 M/s (~1.5×)**. End-to-end DOOP is ~flat (interning is parallelized and dwarfed by the join analysis), but it's free and helps load-bound workloads + the mid-dataflow resolve path.

---

## Results — DOOP, `--str-intern`, 32 workers (output is byte-identical)

| run | baseline | optimized | speedup |
|---|---|---|---|
| batik output phase (51.9M rows, 8.6 GB → tmpfs) | 11.0s | **5.1s** | **2.15×** |
| batik total → tmpfs | 20.0s | **14.2s** | 1.41× |
| batik total → disk | 33.2s | **17.2s** | **1.93×** |
| luindex output phase (9.6M rows) | ~2.0s | **~1.0s** | ~2.1× |

## Validation

- Output byte-identical (26 luindex files / 9.6M rows diff-clean; batik 51,877,104 rows; DOOP printsize counts identical SipHash vs FxHash).
- `cargo test` (215) ✓ · `clippy --all-targets -- -D warnings` ✓
- End-to-end fixtures: **109/109 compiler-path + 109/109 lib-path** (batch + incremental) ✓

## Not included (deliberately)

- **Parallelizing the drain** across worker cores is the largest remaining win, but a naive version materializes formatted bytes (RSS blowup by the full output size, ~8.6 GB for batik); a memory-frugal streaming-pipeline version is a substantial standalone change. Worth a follow-up.
- Reworking the interner for lower memory (dropping the `Spur→&str` map) needs a different interner — bigger than low-hanging fruit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -359,6 +359,7 @@ dependencies = [
  "lasso",
  "ordered-float",
  "regex",
+ "rustc-hash",
  "serde",
  "timely",
 ]
@@ -742,6 +743,12 @@ name = "regex-syntax"
 version = "0.8.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dc897dd8d9e8bd1ed8cdad82b5966c3e0ecae09fb1907d58efaa013543185d0a"
+
+[[package]]
+name = "rustc-hash"
+version = "2.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "94300abf3f1ae2e2b8ffb7b58043de3d399c73fa6f4b73826402a5c457614dbe"
 
 [[package]]
 name = "rustix"
--- a/crates/flowlog-build/src/build/imports.rs
+++ b/crates/flowlog-build/src/build/imports.rs
@@ -146,5 +146,9 @@ fn string_intern_imports(f: &Features) -> TokenStream {
         .string_resolve()
         .then(|| quote! { use ::flowlog_runtime::intern::resolve; });
 
-    quote! { #base #resolve }
+    let resolve_out = f
+        .string_resolve_out()
+        .then(|| quote! { use ::flowlog_runtime::intern::resolve_out; });
+
+    quote! { #base #resolve #resolve_out }
 }
--- a/crates/flowlog-build/src/build/relation/user.rs
+++ b/crates/flowlog-build/src/build/relation/user.rs
@@ -65,7 +65,8 @@ pub(crate) fn user_to_tuple_expr(
 }
 
 /// Inverse of `user_to_tuple_expr`: internal slot → user-facing expression,
-/// used at drain time.
+/// used at drain time. Interned strings resolve through the flat snapshot
+/// path (`resolve_out`) since drain runs after the dataflow's fixpoint.
 pub(crate) fn tuple_to_user_expr(
     dt: &DataType,
     string_intern: bool,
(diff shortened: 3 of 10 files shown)


hdz284:
**Comparison with #136 + combined result → #137.**

@HarukiMoriarty's #136 independently built the parallel drain I flagged as future work here. They're complementary on orthogonal axes:

- this PR: lock-free `resolve_out` (70× on the resolve op) + FxHash, but **single-threaded** drain
- #136: **parallel** drain (rayon + concurrent relation drains) + flush() safety, but still the slow `resolve()`

Benchmarked all three on batik (51.9M rows → tmpfs, 3-run avg, output phase):

| | -w32 | -w64 |
|---|---|---|
| #135 (this, single-thread) | 5.22s | 5.29s |
| #136 (parallel, `resolve()`) | 3.27s | 3.44s |
| combo (#137) | 3.22s | 3.24s |

Parallelism is the dominant output win (1.6× over this PR's single-threaded resolver). And combining the two is strictly better than either: #136's `resolve()` contends a `DashMap`/`RwLock` across worker threads, so it *degrades* 3.27→3.44s from 32→64 workers; swapping in `resolve_out` makes it lock-free and flat (3.22→3.24s).

I opened **#137** = #136 + this PR's two commits (`resolve_out` + FxHash) cherry-picked on top. Suggested landing: merge #136 first, then this PR trims to exactly those two commits — same tree as #137.

PR #136, opened 2026-06-08
perf: parallel output drain

HarukiMoriarty:
## What

Speeds up binary-mode `.output` file writing, which on DOOP-scale runs was the dominant cost after the dataflow solve. Two commits, both confined to output codegen:

- **`fix(codegen): flush file sinks explicitly`** — generated drains relied on `BufWriter::Drop` for the final flush, which silently swallows errors. The last ≤8 KiB of every `.output` file (and the whole single-line `.printsize` file) could be lost on `ENOSPC`/IO error while the binary still exited 0. Now each file sink ends with `out.flush().expect(...)` (threaded through `gen_drain_block` as a `sink_postamble` so the ORDER BY/LIMIT arms get it too). Demonstrated: writing to a full device now panics with non-zero exit where it previously exited 0 with data silently lost.

- **`perf(codegen): parallelize binary .output file writes`** — the real optimization (below).

## The optimization

The old `.output` drain ran entirely on worker 0 after the barrier: per-row `writeln!` through an 8 KiB `BufWriter`, one relation at a time, while the other 63 workers sat idle. Three changes:

1. **Parallel formatting.** File sinks with no `ORDER BY` and arity > 0 now format each *per-worker buffer* into bytes with `rayon` and write them in worker order — so the file is byte-for-byte identical. Parallelism follows the existing `-w N` partitioning; there are **no tuning constants** (no segment size, no wave size, no thread count). Rows are assembled as raw bytes: `resolve()`d strings copied verbatim, integers via `itoa`, floats/bool via `core::fmt` to preserve exact text.

2. **Concurrent relation drains.** The merge section runs each relation's drain in its own scoped thread. Distinct files → distinct inode locks, so large relations write in parallel instead of serializing (this is what overlaps `VarPointsTo` and `InstanceFieldPointsTo` on DOOP).

3. **Plumbing.** `itoa`/`rayon` added as feature-gated scaffold deps (emitted fully qualified, no `use`, so `-Dwarnings` stays clean). The parallel-path predicate is a single `Relation::uses_parallel_file_drain` shared by the drain router and the feature marking so they can't drift.

### Untouched on purpose
Nullary outputs (`True` line), `ORDER BY`/`LIMIT` (global sort), and stderr (`-D -`, one shared stream) keep the sequential drain. Library mode is unaffected.

## Results

Measured end-to-end on `example/program_analysis/doop.dl` over the batik DaCapo corpus (`--str-intern`, `-w 64`, 9.35 GB across 26 outputs, tmpfs), `/usr/bin/time` wall clock:

| | Total wall | Output write phase |
| --- | --- | --- |
| before | ~25.1 s | ~13.3 s |
| after | ~15.1 s | ~3.6 s |

**~1.66× end-to-end; ~3.7× on the write phase.** Output is byte-identical — total bytes match exactly and all 26 relation row counts match.

## Tests

- New fixtures `output_all_types` and `output_all_types_intern` pin byte-exact formatting across all 12 data types, a nullary `.output`, a derived-empty relation, and both string-intern modes. Goldens were generated by the pre-change sequential compiler, so they assert the parallel path reproduces the old bytes.
- Full suite green: 271 workspace unit tests, all 111 binary fixtures (incl. all 18 incremental), and library-mode fixtures.

## Tradeoff

Dropping the segment/wave bounding means formatted bytes are held in memory before writing (~one copy of the output) rather than streamed. On the test machine this is immaterial; on a memory-tight host with a very large output it raises transient peak RSS. A bound can be reintroduced behind the scenes if it ever matters.


hdz284:
This is the right output architecture — parallelism is the dominant win (1.6× over the single-threaded flat-resolver approach in #135). One free stack on top: the parallel path still calls `resolve()` (a `DashMap` lookup hashing the key + `RwLock` read guard), so it becomes a *contention* point under 64-way parallelism.

Benchmark on batik (51.9M rows → tmpfs, 3-run avg, output phase):

| | -w32 | -w64 |
|---|---|---|
| this PR (`resolve()`) | 3.27s | 3.44s |
| + #135's `resolve_out` (combo) | 3.22s | 3.24s |

This PR *degrades* 3.27→3.44s going 32→64 workers (more threads contending the resolve `DashMap`); routing the same parallel drain through #135's flat `Spur`-indexed `resolve_out` makes it lock-free and flat. I combined them in **#137** (this PR + #135's `resolve_out`/FxHash commits cherry-picked); output is set-equal across all 26 batik relations. Nothing to change here — just flagging the complementary win.

PR #137, opened 2026-06-08
perf: faster, leaner string `.output` (combines #135 + #136)

hdz284:
Makes string `.output` **fast and memory-light**. Combines #135 + #136 and fixes the one problem each had on its own.

## The problem

With `--str-intern`, FlowLog stores each string once and uses a small integer key (`Spur`) in its place — great for joins, but at output time every key has to be turned back into its text.

The old drain did that the expensive way: each lookup went through `ThreadedRodeo::resolve`, a concurrent hash map that **hashes the key and grabs a lock every single time**. Multiply that by tens of millions of cells, all on **one thread** (worker 0), and writing the output took longer than computing it.

## The fix (3 parts)

- **Cheap lookups** (from #135). Interned keys are just dense integers `0..N`, so instead of a hashed+locked map we resolve them through a plain array indexed by the key — O(1), no hashing, no lock. About **70× faster per lookup**. (Built once after the computation finishes, so there's nothing to race.)

- **Parallel writes** (from #136). The output is already split into one buffer per worker, so we format those buffers across all cores with `rayon`, and let each relation's file write run concurrently. No more single-thread bottleneck.

  Why both halves matter: #136 alone kept the *slow* locked lookup, so adding threads just made them **fight over the same lock** — it actually got *slower* from 32→64 cores. With the lock-free lookup underneath, the parallel version scales cleanly instead. The two changes only pay off together.

- **Bounded memory** (new). The obvious way to write in parallel is "format everything into memory, then dump it" — but that holds a second full copy of the output (≈15 GB on the test below). Instead we go **wave by wave**: format a handful of fixed-size row chunks at a time into a small pool of **reused** buffers, write them out in order, then reuse them for the next wave. Peak memory is one wave's worth (a few hundred MB), no matter how big the output gets — so even many relations writing at once stay safe.

Plus a small correctness fix (from #136): flush files explicitly, so a failed final write surfaces as an error instead of being silently dropped.

## Results

36M rows / 7.3 GB string output, `--str-intern`, `-w 64`:

| | speed | peak memory |
|---|---|---|
| before (`main-next`) | 10.6s | 2.6 GB |
| #135 + #136, naive combine | 5.6s | **~15 GB** 😱 |
| **this PR** | **5.3s** | **2.6 GB** ✅ |

**~2× faster, same memory as before** (and up to ~4× faster when many relations write at once). Output is byte-for-byte identical.

## Tested

- 271 unit tests ✓ · `clippy -D warnings` ✓
- All 111 end-to-end fixtures pass on both the compiler and library paths (batch + incremental).
- Stress-tested vs. the baseline — segment boundaries (8191/8192/8193 rows…), new strings appearing across incremental epochs, all 12 column types, wide rows, many concurrent relations, and awkward strings (unicode, embedded delimiters, empty). All byte-identical.

> One note on checking output: with multiple workers and no `ORDER BY`, row *order* varies run to run (that's just how the workers interleave — Datalog relations are sets). So the suite compares sorted output, or byte-exact at `-w 1`. Both match the baseline exactly.

Credit: parallel drain + flush fix by @HarukiMoriarty (#136); flat resolve + FxHash interner from #135.


hdz284:
Pushed a 5th commit (`9cecfd5`) that swaps the collect-everything step for a **bounded-streaming** writer: format `rayon`-many fixed-size row segments per wave into a reused scratch pool and write each wave straight to a 1 MiB `BufWriter` in worker/row order. This removes the former "hold ≈one copy of the output in RAM" tradeoff while keeping the parallel speed.

End-to-end (36M-row / 7.3 GB string `.output`, `--str-intern`, `-w 64`, tmpfs): **5.3 s wall, 2.67 GB peak RSS** — same speed as the collect path (~5.6 s) at **~5.6× less memory** (~15 GB → 2.67 GB), i.e. the baseline single-thread drain’s footprint. Output byte-identical at `-w 1`, sorted-md5 identical at `-w 64`. 271 unit tests + clippy `-D warnings` + the `output_all_types{,_intern}` / multi-worker / ORDER BY / LIMIT / incremental fixtures all green on both compiler and library paths.

HarukiMoriarty:
Why `PARALLEL_DRAIN_SEG_ROWS` (8192) bounds *memory*, not *speed*. The drain is
bulk-synchronous: each wave formats `L` segments in parallel, then writes them in
order, so the wall-clock is

$$
T = \frac{N}{Lr} + \frac{Nw}{D} + \frac{No}{SL}
$$

a parallel **format** term, a serial **write** term, and a per-wave **overhead**
term, where

| symbol | meaning | unit |
|:--:|---|:--:|
| $N$ | output rows | rows |
| $w$ | bytes per formatted row | bytes/row |
| $B = Nw$ | total output bytes | bytes |
| $L$ | parallel lanes (`rayon` threads ≈ cores) | — |
| $S$ | `PARALLEL_DRAIN_SEG_ROWS` | rows/segment |
| $r$ | per-core format rate (resolve + itoa + memcpy) | rows/s |
| $D$ | sink write bandwidth | bytes/s |
| $o$ | per-wave `rayon` dispatch + join overhead | s |

## Derivation

There are $N/(SL)$ waves; format and write do not overlap, so each wave costs
$S/r + LSw/D + o$. Multiplying through:

$$
T = \frac{N}{SL}\left(\frac{S}{r} + \frac{LSw}{D} + o\right) = \frac{N}{Lr} + \frac{Nw}{D} + \frac{No}{SL}
$$

## What the formula says

**$S$ cancels out of the work.** Total formatting ($N$ rows across $L$ cores) and
total writing ($B$ bytes at $D$) are independent of how rows are chunked — $S$
survives *only* in the overhead term, in the denominator.

**$S$ is off the critical path once** the overhead is small next to the format
term:

$$
\frac{No}{SL} \ll \frac{N}{Lr} \Longleftrightarrow S \gg o r
$$

With dispatch overhead $o$ on the order of a few microseconds and per-core format
rate $r \approx 2 \times 10^{6}$ rows/s, the product $or$ is only ~10–20 rows.
$S = 8192$ clears that by ~400×, so the overhead term is noise.

**The real ceiling.** Dropping the overhead,

$$
T \approx \frac{B}{Lrw} + \frac{B}{D} = \frac{B}{\min(Lrw, D)}
$$

the smaller of aggregate format bandwidth $Lrw$ and disk bandwidth $D$. When
$Lrw < D$ the drain is **format/CPU-bound** (scale cores $L$, or cheapen the
per-row rate $r$); when $Lrw > D$ it is **disk-bound** (overlapping format and
write would help, more cores do not).

## Takeaways

- **$S$ is a memory knob.** Peak in-flight bytes are about $LSw$ per relation
  (one wave, independent of $N$) — tune it for footprint, not speed.
- **Throughput** is pinned to $\min(Lrw, D)$, in which $S$ does not appear.
- On the tmpfs benchmark $D \gg Lrw$, so the drain is **format-bound** — exactly
  what this PR targets: `resolve_out` raises $r$ and `rayon` raises $L$, both
  lifting $Lrw$, the term that was the wall.
- Second-order: a final partial wave adds a one-time tail of about $S/r$ that
  *grows* with $S$; combined with the overhead's $1/S$ falloff, this is the
  shallow U-shaped "knee" that 8192 sits near.

PR #138, opened 2026-06-10
perf(batch): park instead of busy-spin when draining to fixpoint

hdz284:
## TL;DR

Batch programs currently drain their Timely dataflow with `worker.step()`.
Workers with no pending scheduler activity busy-spin while waiting at progress
barriers, consuming a core without advancing the computation.

This PR uses an adaptive batch-only drain:

- After scheduler activity, keep using `worker.step()` for up to 4096
  consecutive idle steps. This avoids park/wake latency on active and
  short-fixpoint runs.
- Once the worker has stayed idle for that window, use
  `worker.step_or_park(Some(1ms))`.
- New scheduler activity resets the spin window.

Incremental mode is unchanged.

## Current implementation

The policy now lives in the shared batch step-loop generator, so both the
compiler and `flowlog-build` library paths use it. It also preserves periodic
profile-metric flushes added on `main-next`.

The activation handle is captured once outside the loop. The active path still
has a predictable threshold branch and an `empty_for()` check, but it no longer
clones the activation `Rc` on every scheduler step.

### Large recursive workloads

Adaptive spin-then-park retains the large CPU reduction that motivated the PR:

| program / dataset / workers | baseline wall / CPU | adaptive wall / CPU |
|---|---:|---:|
| ctx / xalan / w16 | 36.5 s / 581 s | 36.4 s / 200 s |
| ctx / xalan / w32 | 38.2 s / 1210 s | 37.0 s / 389 s |
| ctx / h2o / w32 | 388 s / 12367 s | 377.6 s / 2206 s |

That is 66-82% less total CPU with flat or better wall time in the target
workloads.

### Shorter DOOP matrix

Across the reported 33 cells (11 DaCapo suites at w8/w16/w32):

- Mean CPU: -10.4%.
- Mean wall: +1.7%.
- Repeated measurements of the noisy larger w32 cases showed a residual
  3-5% wall increase, paired with roughly 14-20% lower CPU.

This is a CPU/energy versus latency tradeoff, not an unconditional wall-time
win.

### Deep, narrow recursion

After rebasing onto current `main-next`, I also measured a deliberately narrow
recursion:

```souffle
.decl Reach(x: int32)
Reach(0).
Reach(x + 1) :- Reach(x), x < 500000.
.printsize Reach
```

At w1 over 15 paired warm runs:

- Baseline median: 312.866 ms.
- Adaptive median: 314.283 ms (+0.45%).
- The means differed by -0.25%, so the effect is within run-to-run noise.

At w8, paired warm runs were also flat (adaptive medians were about 0.8-0.9%
lower for both wall and CPU). Instrumentation observed roughly 25 million
scheduler calls per worker but only 24-42 park calls. Progress traffic therefore
keeps this workload on the spin path: the adaptive branch does not materially
hurt latency, but it also does not reclaim much CPU in this particular shape.

## Correctness and validation

Changing the worker scheduling policy does not change the Datalog fixpoint.
The original branch passed the compiler-mode Souffle oracle suite and produced
identical relation results in the benchmark matrix.

The rebased commit additionally passes:

- `cargo +nightly fmt --all --check`
- `cargo test`
- Compiler fixtures: `recursive_tc`, `recursive_intermediate`
- Library fixtures: `recursive_tc`, `recursive_intermediate`
- A profiled generated run with a 1 ms metrics-flush interval

The commit is signed off and the incremental drain remains unchanged.

### DIFF
--- a/flowlog-build/src/codegen/profile.rs
+++ b/flowlog-build/src/codegen/profile.rs
@@ -280,11 +280,39 @@ impl CodeGen {
     /// Emits the batch run loop (`worker` and `index` in scope): steps the
     /// dataflow to fixpoint, periodically flushing metrics when profiled.
     pub(crate) fn gen_batch_step_loop(&self) -> TokenStream {
-        self.gen_flush_loop(
-            quote! { worker.step() },
-            quote! {},
-            self.gen_metrics_write_batch(),
-        )
+        let step = quote! {{
+            let more = if __flowlog_idle_steps < __FLOWLOG_SPIN_BEFORE_PARK {
+                worker.step()
+            } else {
+                worker.step_or_park(Some(__FLOWLOG_PARK_TIMEOUT))
+            };
+            if more {
+                if __flowlog_activations.borrow().empty_for()
+                    == Some(std::time::Duration::ZERO)
+                {
+                    __flowlog_idle_steps = 0;
+                } else if __flowlog_idle_steps < __FLOWLOG_SPIN_BEFORE_PARK {
+                    __flowlog_idle_steps += 1;
+                }
+            }
+            more
+        }};
+        let step_loop = self.gen_flush_loop(step, quote! {}, self.gen_metrics_write_batch());
+
+        quote! {
+            {
+                // Spin briefly after scheduled work to avoid a park/wake round
+                // trip at progress barriers. Once idle, park with a bounded
+                // timeout so workers stop consuming a core while profiling can
+                // still flush on schedule.
+                const __FLOWLOG_SPIN_BEFORE_PARK: u64 = 4096;
+                const __FLOWLOG_PARK_TIMEOUT: std::time::Duration =
+                    std::time::Duration::from_millis(1);
+                let __flowlog_activations = worker.activations();
+                let mut __flowlog_idle_steps = 0u64;
+                #step_loop
+            }
+        }
     }
 
     /// Emits the incremental commit step loop (`worker`, `probe`,


HarukiMoriarty:
Though it looks like a "dark magic", I highly suspect at some situations (e.g. incremental) the performance will degraded. I personally test on batch mode, seems it also help with fewer threads, before we merge this tricks, we need to do a more comprehensive investigation.

hdz284:
## Re-investigation + a cleaner fix that removes the wall regression

I dug into @HarukiMoriarty's concern (regression on short fixpoints / few threads, "needs comprehensive investigation"). Short version: **the gap is real, `step_or_park(None)` does fix the CPU but the wall regression is real too — and an adaptive spin-then-park gets the same CPU win with *no* wall regression at any worker count.**

All runs below: base `c1e8841`, `--str-intern --mode datalog-batch`, `/usr/bin/time -v`, on the 20-DaCapo fact DBs. Three drain variants, same engine otherwise:

- **V0** = `while worker.step() {}` (today's busy-spin baseline)
- **V1** = `while worker.step_or_park(None) {}` (this PR)
- **V3** = adaptive spin-then-park (proposed; diff at the bottom)

### 1. The gap is real
ctx-insensitive (flattened DOOP) busy-spins idle workers at 100% CPU:

| | wall | total CPU |
|---|--:|--:|
| xalan w16 | 36.5 s | 581 s |
| xalan w32 | 38.2 s | **1210 s** (≈32 cores pinned; wall *no better* than w16) |

So the extra 16 workers at w32 add ~615 s of pure spin for **zero** wall benefit. Confirmed.

### 2. `step_or_park(None)` (V1) fixes CPU but regresses wall — not just on short fixpoints
The per-round park→wake latency lands on the critical path. It shows up on the *long* recursive program too, not only short ones:

| program · dataset · w | V0 wall / CPU | V1 wall / CPU |
|---|--:|--:|
| ctx · xalan · w16 | 36.5 s / 581 s | 39.0 s / 190 s → **+6.7 % wall**, −67 % CPU |
| ctx · xalan · w32 | 38.2 s / 1210 s | 40.5 s / 361 s → **+5.9 % wall**, −70 % CPU |
| ctx · h2o · w32 | 388 s / 12367 s | 404 s / 1620 s → **+4.1 % wall**, −87 % CPU |
| doop · xalan · w32 | 2.7 s / 82 s | 3.1 s / 67 s → **+12.5 % wall**, −18 % CPU |
| doop · xalan · w8 | 3.9 s / 30 s | 4.0 s / 26 s → +3.6 % wall, −13 % CPU |

So the reviewer's instinct was right: V1 trades wall for CPU, and the trade is worst exactly where the engine is already fast.

### 3. Adaptive spin-then-park (V3): same CPU win, no wall regression
Idea: `worker.step()` is literally `step_or_park(Some(0))` (never parks), and `activations().empty_for() == Some(0)` is timely's own "work pending now" signal. So **spin (zero-timeout) while the worker has recently had work, and only park once it's been idle for a sustained stretch.** Short fixpoints stay hot (no park/unpark latency); long fixpoints park their idle workers and reclaim the cores.

| program · dataset · w | V0 wall / CPU | **V3 wall / CPU** | vs V1 |
|---|--:|--:|--:|
| ctx · xalan · w16 | 36.5 s / 581 s | **36.4 s / 200 s** (−0.5 % wall, **−66 % CPU**) | same CPU, no regression |
| ctx · xalan · w32 | 38.2 s / 1210 s | **37.0 s / 389 s** (−3.2 % wall, **−68 % CPU**) | same CPU, **faster than V0** |
| ctx · h2o · w32 | 388 s / 12367 s | **377.6 s / 2206 s** (−2.6 % wall, **−82 % CPU**) | V1 was +4.1 % wall / 1620 s; V3 **faster than V0** |
| doop · xalan · w32 | 2.7 s / 82 s | **2.8 s / 78 s** (+1.8 % wall) | regression gone (was +12.5 %) |
| doop · h2o · w32 | 15.6 s / 464 s | **15.2 s / 378 s** (−2.8 % wall, −19 % CPU) | better on both |
| doop · xalan · w8 | 3.9 s / 30 s | **3.9 s / 29 s** (+0.5 % wall) | regression gone (was +3.6 %) |

**V3 dominates both:** vs V0 it cuts CPU 66–68 % with equal-or-better wall; vs V1 it keeps the CPU win but removes the wall regression everywhere (short *and* long fixpoints, w8/w16/w32).

The knob isn't knife-edge — sweeping `FLOWLOG_SPIN` on ctx-xalan w32 (spin count before a worker parks):

| spin | wall | CPU |
|--:|--:|--:|
| 0 (≈always park, like V1) | 39.1 s | 373 s |
| 1024 | 37.7 s | 389 s |
| **4096 (default)** | **37.2 s** | **396 s** |
| 16384 | 36.9 s | 398 s |
| 262144 | 36.9 s | 406 s |

Anywhere from ~1k–260k gives wall ≤ V0 with CPU ~390–406 s (vs V0's 1210 s). 4096 is a comfortable default.

### Correctness & safety
- **Bit-identical results**: per-relation cardinalities match across V0/V1/V3 (doop 26 relations, ctx-insensitive 20 relations, 0 mismatches).
- **Scoped to the batch drain only** (`gen_batch_main`) — incremental mode is untouched, so the `datalog-inc` concern doesn't apply.
- **Safer than V1**: the park is *bounded* (`Some(1ms)`), so even a missed buzz can't wedge the drain — it self-rechecks; `None` parks indefinitely.

### Proposed diff (production form — drop the env knob, fixed 4096)
```rust
// Drain to fixpoint with adaptive spin-then-park. `worker.step()` is
// `step_or_park(Some(0))` (never parks); spin while the worker has recently
// had work so short fixpoints never pay park/unpark latency, then park once
// idle for SPIN_BEFORE_PARK steps so large recursive programs stop burning
// whole cores on the progress-barrier spin. `empty_for() == Some(0)` is
// timely's own "work pending now" signal.
const SPIN_BEFORE_PARK: u64 = 4096;
let mut idle_steps: u64 = 0;
loop {
    let more = if idle_steps < SPIN_BEFORE_PARK {
        worker.step()
    } else {
        worker.step_or_park(Some(Duration::from_millis(1)))
    };
    if !more { break; }
    if worker.activations().borrow().empty_for() == Some(Duration::ZERO) {
        idle_steps = 0;
    } else {
        idle_steps = idle_steps.saturating_add(1);
    }
}
```

Happy to push this onto the branch if we like the approach. (I benchmarked it via a `FLOWLOG_SPIN` env override so one binary could sweep the whole V0↔V1 spectrum; the production form just fixes 4096.)

HarukiMoriarty:
Thanks for the re-investigation — V3 (adaptive spin-then-park) looks like the right direction: same CPU win as V1, no wall regression, and the bounded `Some(1ms)` park is safer than `None`.

Before we take it, two asks:

1. **Please run `doop.dl` on the full DaCapo — all 20 suites — and report performance**: wall + CPU (`/usr/bin/time -v`) at the worker counts you've been testing (w8/w16/w32), V0 vs V3. The tables so far only show xalan and h2o for doop, and doop is exactly the short-fixpoint regime where V1 regressed — we should see the no-regression result hold across all 20 suites before hardcoding `SPIN_BEFORE_PARK = 4096`.

2. Push the production form of V3 to this branch (fixed const, no env knob) and re-run the Soufflé-oracle suite in compiler mode against that exact diff, same as you did for V1.

Note: I've rebased this branch onto the new repo layout (`crates/` → repo root), so reset your local branch to `origin/perf/batch-step-or-park` before pushing on top.

hdz284:
Pushed the production form of V3 (`a77c958`, fixed `SPIN_BEFORE_PARK = 4096`, no env knob, signed off) and ran the two asks below. Reset onto `origin/perf/batch-step-or-park` first, as you noted.

## 1. `doop.dl` across DaCapo — V0 (`step()`) vs V3 (adaptive), wall + CPU

Harness: `doop/default.dl` compiled `--str-intern --mode datalog-batch` with each ref's compiler, `/usr/bin/time -v`, page cache warmed per suite so both variants see warm facts. All locally-cached DaCapo suites (11); results size-report-identical V0↔V3 on every one (parity below). `jython` excluded — it's the long-fixpoint outlier (~8 min/run, 32 GB RSS), i.e. the regime *most* favorable to V3; I didn't want to hold the reply on its full 6-run matrix.

**-w 8**

| suite | wall V0 | wall V3 | Δwall | CPU V0 | CPU V3 | ΔCPU |
|---|--:|--:|--:|--:|--:|--:|
| tomcat | 3.8s | 3.8s | +1.1% | 30s | 28s | -7.7% |
| eclipse | 17.9s | 18.0s | +0.7% | 142s | 119s | -16.0% |
| luindex | 4.5s | 4.5s | -0.2% | 35s | 32s | -8.8% |
| lusearch | 3.0s | 3.0s | +0.0% | 24s | 22s | -9.8% |
| sunflow | 7.8s | 8.1s | +3.3% | 62s | 55s | -10.8% |
| avrora | 2.9s | 3.0s | +1.7% | 23s | 22s | -2.7% |
| fop | 18.8s | 18.9s | +1.0% | 149s | 129s | -13.2% |
| pmd | 5.9s | 5.9s | -0.2% | 47s | 43s | -7.8% |
| h2 | 18.1s | 18.2s | +0.4% | 144s | 127s | -11.9% |
| batik | 19.2s | 19.4s | +0.9% | 152s | 134s | -12.3% |
| xalan | 3.9s | 3.9s | -0.3% | 30s | 29s | -4.7% |

**-w 16**

| suite | wall V0 | wall V3 | Δwall | CPU V0 | CPU V3 | ΔCPU |
|---|--:|--:|--:|--:|--:|--:|
| tomcat | 2.6s | 2.7s | +2.7% | 41s | 39s | -4.6% |
| eclipse | 12.2s | 12.1s | -0.7% | 193s | 156s | -19.5% |
| luindex | 3.2s | 3.2s | +0.6% | 50s | 45s | -9.2% |
| lusearch | 2.3s | 2.3s | +0.9% | 36s | 31s | -12.7% |
| sunflow | 5.4s | 5.4s | +0.4% | 84s | 74s | -11.9% |
| avrora | 2.1s | 2.2s | +1.9% | 33s | 32s | -3.6% |
| fop | 12.4s | 12.6s | +2.2% | 195s | 167s | -14.5% |
| pmd | 4.2s | 4.2s | -0.2% | 65s | 58s | -9.6% |
| h2 | 11.7s | 11.8s | +1.2% | 184s | 160s | -13.4% |
| batik | 12.8s | 12.9s | +0.2% | 202s | 174s | -14.0% |
| xalan | 2.7s | 2.8s | +1.1% | 42s | 41s | -3.4% |

**-w 32**

| suite | wall V0 | wall V3 | Δwall | CPU V0 | CPU V3 | ΔCPU |
|---|--:|--:|--:|--:|--:|--:|
| tomcat | 2.3s | 2.3s | +1.3% | 71s | 67s | -4.8% |
| eclipse | 9.3s | 10.5s | +13.0%¹ | 293s | 252s | -13.9% |
| luindex | 2.7s | 2.7s | +2.6% | 82s | 74s | -9.6% |
| lusearch | 2.3s | 2.2s | -3.5% | 70s | 57s | -18.5% |
| sunflow | 4.3s | 4.6s | +5.1% | 135s | 122s | -9.3% |
| avrora | 2.1s | 2.1s | -2.3% | 64s | 58s | -9.5% |
| fop | 9.2s | 10.0s | +8.9%¹ | 285s | 254s | -10.9% |
| pmd | 3.4s | 3.6s | +3.8% | 105s | 95s | -9.5% |
| h2 | 8.5s | 8.6s | +1.5% | 266s | 228s | -14.2% |
| batik | 9.6s | 10.3s | +7.2%¹ | 300s | 266s | -11.4% |
| xalan | 2.5s | 2.6s | +1.6% | 76s | 70s | -8.6% |

**Mean over 33 cells: ΔCPU −10.4%, Δwall +1.7%.**

### Honest read (not "zero regression")

- **CPU is down on every single cell** (−2.7% … −19.5%), and the win grows with worker count — exactly the idle-worker spin that V1 also removed. This is the target metric and it's unambiguous.
- **Wall is flat within noise on the short/small suites** (tomcat, xalan, avrora, lusearch, luindex, pmd — all ≤~4 s), which is precisely the short-fixpoint regime where V1 regressed +3–12%. V3 does **not** regress there. That's the result you wanted to confirm, and it holds.
- **The bigger suites at -w32 do carry a small residual wall cost.** ¹ Those three cells are single-run; I re-measured them 5× (warm, medians) because the eclipse +13% looked like an outlier — it was:

  | cell | V0 wall (med of 5) | V3 wall (med of 5) | Δwall | ΔCPU |
  |---|--:|--:|--:|--:|
  | eclipse -w32 | 9.26s | 9.51s | **+2.7%** | −19.9% |
  | fop -w32 | 9.11s | 9.54s | **+4.7%** | −13.9% |
  | batik -w32 | 9.56s | 9.86s | **+3.1%** | −14.0% |

  So the real w32 big-suite wall regression is ~**+3–5%** (the +13% was variance), bought back by ~14–20% CPU. Still park-wakeup latency on the critical path, just an order smaller than V1's, and only on the large-program × max-worker corner.

Net: V3 keeps V1's CPU/energy win, erases V1's short-fixpoint wall regression, and shrinks the residual to ≤~5% wall on big programs at w32. `SPIN_BEFORE_PARK` is the lever there — pushing it higher trims that last few-percent at some CPU cost — but I left it at 4096 per your "fixed const" ask rather than tuning to the corner. Happy to bump it if you'd rather buy back the w32 big-suite wall.

## 2. Soufflé-oracle suite, compiler mode, on this exact diff

`tests/oracle/run_compiler.sh` against `a77c958`:

```
[PASS] tc (24730729 tuples)
[PASS] sg (24730729 tuples)
[FINISH] All tests passed
```

Both are recursive fixpoints, so they exercise the new drain loop directly. Plus, since a drain schedule can't change the fixpoint, I diffed V0 vs V3 `.printsize` reports on all 11 DaCapo datasets above — **identical on every relation, every suite** — and TC output is bit-identical at -w1/-w4. V0 is `main-next`, already oracle-validated against Soufflé, so V0≡V3 ⇒ V3 matches Soufflé too.

HarukiMoriarty (approved):
Rebased onto current main-next, shared batch-loop integration verified, critical-path overhead measured, and full CI is green. The CPU/wall tradeoff is documented.

HarukiMoriarty:
Note: this change was removed from `main-next` on 2026-08-08 due to a performance regression (wall-time cost in batch runs — the +1.7% mean / 3-5% residual at 32 workers noted above turned out to matter more than the CPU savings).

Per the branch policy, `main-next` history was rewritten to drop the commit rather than adding a revert, so `35a619c` is no longer reachable from the branch; stacked PRs were rebased.

The CPU reclaim idea itself is still welcome — happy to see it re-proposed with different tuning (spin threshold / park timeout) if the wall-time cost can be held at zero.

PR #139, opened 2026-06-10
perf: share content-identical arrangements across rules (codegen CSE)

hdz284:
## What

A codegen-level arrangement-sharing (CSE) pass: when two rules build a
content-identical arrangement, the second reuses the first's projected
collection and sorted trace instead of rebuilding them.

## Why

The planner dedups arrangements by their **lineage-bearing** output fingerprint,
which encodes rule-variable *slot positions*. So the same relation arranged the
same way in two differently-shaped rules produces two separate arrangements even
though the generated dataflow code is identical. On context-insensitive DOOP
this leaves ~100 redundant arrangements per program.

## Example

Two rules that both need `Edge` arranged by its first column:

```datalog
Reach(x, y) :- Seed(x),    Edge(x, y).   // Edge is the 2nd body atom
Step(x, y)  :- Edge(x, y), Live(x).      // Edge is the 1st body atom
```

Both compile to the same operator — `edge.arrange_by_key()`, keyed on column 0.
But the planner's lineage fingerprint encodes each variable's **rule-slot
position** (`Edge` is the 2nd atom in `Reach`, the 1st in `Step`), so it treats
them as distinct and builds the arrangement **twice**. With this PR they collide
on the generated-code fingerprint, so `Edge` is sorted **once** and the second
rule reuses the trace (a cheap dataflow-handle clone).

## How

`emit_shared_arrangement` keys a per-scope cache on `arrangement_share_key` =
fingerprint of the **generated build tokens** (output ident neutralized) plus
key-only-ness. On a hit it emits `let out = canonical.clone();` (a cheap
dataflow-handle clone) and points the output fingerprint at the already-built
shared trace.

Hashing the rendered code is the tightest *safe* key — codegen is the canonical
normal form of a `TransformationFlow`. It captures content-identical
arrangements that an `(input, flow)` key over-splits (e.g. joins differing only
in codegen-irrelevant `Vec` ordering), while only ever merging byte-identical —
hence type-identical — code. An audit confirmed it reaches the content-optimal
arrangement count (0 missed, 0 over-merge) on both `doop.dl` and
`context-insensitive`.

## Results — baseline vs this PR (32 workers, interleaved, str-intern)

**context-insensitive** — peak RSS win scales with problem size:

| app | VarPointsTo | peak RSS | wall |
|-----|------------:|---------:|-----:|
| avrora  |  2.5M | **−10.0%** | −2.6% |
| xalan   |  2.4M | **−11.1%** | −6.6% |
| luindex |  5.4M |  **−8.7%** | −4.2% |
| h2      | 22.7M | **−15.5%** | −6.3% |
| batik   | 28.2M | **−16.0%** | −1.2% |

(batik and h2 each save ~2.1 GiB of peak RSS.)

**doop.dl** — flat (within noise) on all apps; it has few shareable arrangements.

## Correctness

- `doop.dl`: outputs **byte-exact** vs baseline on all apps.
- `context-insensitive`: identical per-relation counts and `confirm_benign`
  **residual 0** — i.e. it differs only by the pre-existing benign `ord()`
  heap-merge representative renaming, which the baseline itself exhibits
  run-to-run under string interning.
- `cargo clippy -p flowlog-build`: 0 warnings; `cargo test -p flowlog-build`: 271 pass.

### DIFF
--- a/crates/flowlog-build/src/codegen/flow/non_recursive.rs
+++ b/crates/flowlog-build/src/codegen/flow/non_recursive.rs
@@ -40,18 +40,21 @@ impl CodeGen {
         let mut flows = Vec::new();
         let global_fp_to_ident = self.global_fp_to_ident.clone();
         let mut outer_arranged = mem::take(&mut self.outer_arranged);
+        let mut outer_sem_arranged = mem::take(&mut self.outer_sem_arranged);
 
         for transformation in stratum.non_recursive_transformations() {
             flows.push(self.gen_transformation(
                 &global_fp_to_ident,
                 transformation,
                 &mut outer_arranged,
+                &mut outer_sem_arranged,
                 stratum,
                 profiler,
             )?);
         }
 
         self.outer_arranged = outer_arranged;
+        self.outer_sem_arranged = outer_sem_arranged;
 
         trace!("Generated static flows:\n{}\n", quote! { #(#flows)* });
         Ok(flows)
--- a/crates/flowlog-build/src/codegen/flow/recursive.rs
+++ b/crates/flowlog-build/src/codegen/flow/recursive.rs
@@ -113,11 +113,22 @@ impl CodeGen {
         current.extend(recursive_bindings.clone());
 
         // --- Rule transformations -------------------------------------------
+        // Per-block semantic arrangement cache: idents created here live inside
+        // this `iterative(|inner| …)` closure, so sharing must be scoped to the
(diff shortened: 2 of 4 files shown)


hdz284:
Consolidated into #142, which keeps the same content-sharing goal but resolves it entirely in the planner (codegen renders the alias) rather than hashing rendered tokens. It reaches a comparable arrangement count and is verified self-join-sound. Suggest closing this in favor of #142.

PR #141, opened 2026-06-11
perf(planner): share content-identical arrangements at plan time

hdz284:
## Summary

FlowLog under-shares differential-dataflow **arrangements** across rules: the
same relation, projected and keyed identically, is sorted and indexed multiple
times. On wide program-analysis workloads (DOOP / context-insensitive points-to)
a handful of relations are joined across hundreds of rules, so the duplication
adds up — every redundant arrangement is an extra sorted index rebuilt on every
recursive round.

This PR fixes it **entirely in the query plan**: it makes the transformation
fingerprint *content-canonical* so the planner's existing dedup passes collapse
the duplicates. **No codegen changes** (0 files touched under `codegen/`).

## Root cause

`output_info_fp` — the identity the planner's dedup keys on — is hashed over each
transformation's input fingerprint(s), key/value layouts, and predicates. Those
reference columns via `AtomArgumentSignature`, which embeds two **rule-local**
bits that have nothing to do with a collection's *content*:

- `rhs_id` — the atom's position in the rule body, and
- `is_positive` — positive (`R(x)`) vs negative (`!R(x)`) occurrence.

So the *same* logical arrangement (relation `R` keyed on the same columns) gets a
*different* fingerprint across rules — when `R` sits at a different body index,
or is joined positively in one rule and used as an antijoin in another. The two
dedup passes (`deduplicate_transformation_infos` within a stratum,
`prune_cross_stratum_duplicates` across strata) miss it, and codegen emits the
same `arrange_by_key` again.

## Fix

A new `planner/transformation/canonical.rs` produces the operands the fingerprint
is hashed from, with the rule-local noise normalized away:

- every distinct atom is renamed to a canonical index by first-appearance order
  (preserves `argument_id` = *which column*; normalizes `rhs_id` + polarity);
- conjunctive predicate lists are order-normalized (sorted by element fingerprint).

The existing dedup passes then collapse content-identical transformations — the
plan becomes a DAG with shared nodes and codegen lowers it 1:1.

**Sound.** Interning keys on the *full* `AtomSignature`, so the renaming is
injective and never merges genuinely different transformations; canonical
operands are used *only* for the hash (stored layouts/predicates are untouched,
so per-transform codegen output is unchanged); input fingerprints are already
canonical, so it composes bottom-up. Polarity is content-irrelevant because an
arrangement holds the projected relation data — the positive/negative
distinction lives in the *consumer* (semijoin vs antijoin), both of which want
the identical keyed trace.

**Why plan-time (not codegen).** Arrangements are scope-local in differential
dataflow, and the plan already models the scopes: all non-recursive work shares
the one root dataflow scope (deduped program-wide by the cross-stratum prune),
and each recursive stratum is its own `iterative` scope (deduped per-stratum). So
sharing is resolved correctly in the plan, with the plan as the single source of
truth (visible to the profiler / optimizer), and duplicate *projections* are
removed too — not just the `arrange`.

## Results (batik, `--str-intern`, interleaved A/B vs `main-next`)

**Benchmark (interleaved A/B vs `main-next`):**

```
Arrangements emitted — DOOP-135   (fewer = more sharing)
  main-next  ██████████████████████████████████ 204
  this PR    ████████████████████████████████ 192  −5.9%

Arrangements emitted — context-insensitive (592 rules)   (fewer = more sharing)
  main-next  ██████████████████████████████████ 1290
  this PR    ██████████████████████████████▊ 1167  −9.5%

Peak RSS — context-insensitive (-w8)   (lower is better)
  main-next  ██████████████████████████████████ 11.33 GB
  this PR    ███████████████████████████▉  9.27 GB  −18.2%

Wall clock — context-insensitive (-w8)   (lower is better)
  main-next  ██████████████████████████████████ 78.3 s
  this PR    ██████████████████████████████▌ 70.1 s  −10.4%
```

| Program | metric | `main-next` | this PR | Δ |
| --- | --- | ---: | ---: | ---: |
| DOOP-135 | arrangements | 204 | 192 | **−5.9%** |
| DOOP-135 | peak RSS (-w8) | 4214 MB | 4119 MB | −2.3% |
| DOOP-135 | wall (-w8) | 19.6 s | 19.4 s | ~flat |
| context-insensitive (592) | arrangements | 1290 | 1167 | **−9.5%** |
| context-insensitive (592) | peak RSS (-w8) | 11.33 GB | 9.27 GB | **−18.2%** |
| context-insensitive (592) | wall (-w8) | 78.3 s | 70.1 s | **−10.4%** |

DOOP-135 is small (only 12 tiny EDB-projection duplicates), so the win is
within noise there; the payoff scales with program size — on the 592-rule
context-insensitive analysis it is a ~18% peak-RSS reduction and ~10% wall
speedup.

## Correctness

Validated against a fresh `main-next` baseline build on batik: **every output
relation cardinality is identical** — DOOP 26/26 and context-insensitive 20/20
(e.g. `VarPointsTo` = 28,249,074, `InstanceFieldPointsTo` = 4,124,124). For the
points-to relations the multiset is identical; only the ord()-based heap
representatives differ, which is a pre-existing parallel-execution
non-determinism unrelated to this change. Full workspace test suite green; new
unit tests cover the canonicalization (body-position / polarity / predicate-order
invariance) and the soundness guards (distinct columns, distinct atoms, value
origin must not merge); clippy clean.

## Completeness & relation to the codegen approach

A scope-aware audit shows **0 residual same-scope duplicate arrangements** on
both programs. This is the planner-side counterpart to the codegen-CSE in #139;
it captures the same opportunities **plus 11 more** on context-insensitive,
because codegen renders conjuncts in `Vec` order and so a rendered-code key
over-splits on predicate ordering — which the structural key normalizes.

The one remaining duplicate flagged on context-insensitive
(`mainAnalysis.CompatibleTypes` vs `basic.SupertypeOf`) is a copy-rule alias:
two *distinct* relations bound to the same collection. That's a relation-inlining
opportunity, not arrangement sharing — out of scope here.

> Note: the change is unconditional (no feature flag). Recommend a full Soufflé
> oracle run before merge, per the repo's correctness gate.

### DIFF
--- a/crates/flowlog-build/src/planner/transformation.rs
+++ b/crates/flowlog-build/src/planner/transformation.rs
@@ -12,6 +12,7 @@ use tracing::trace;
 use crate::catalog::JoinPredicates;
 use crate::planner::Collection;
 
+mod canonical;
 mod flow;
 mod info;
 
--- a/crates/flowlog-build/src/planner/transformation/canonical.rs
+++ b/crates/flowlog-build/src/planner/transformation/canonical.rs
@@ -1,0 +1,412 @@
+//! Content-canonical fingerprints for plan-time arrangement sharing.
+//!
+//! A transformation's `output_info_fp` (see [`super::info`]) is the identity of
+//! the collection it produces. It is hashed from the transformation's input
+//! fingerprint(s), key/value layouts, and predicates — and those layouts and
+//! predicates reference columns via [`AtomArgumentSignature`], which carries two
+//! pieces of *rule-local* metadata that are irrelevant to a collection's
+//! content:
+//!
+//!   - `rhs_id`      — the source atom's position in the rule body, and
+//!   - `is_positive` — whether the occurrence was positive (`R(x)`) or negative
+//!     (`!R(x)`).
+//!
+//! Hashing those splits the fingerprint of the *same* logical arrangement
+//! (relation `R` keyed on the same columns) across rules — e.g. when `R` sits
+//! at a different body index, or is joined positively in one rule and used as an
+//! antijoin in another. The planner's two dedup passes
+//! ([`StratumPlanner::deduplicate_transformation_infos`] within a stratum and
+//! [`prune_cross_stratum_duplicates`] across strata) both key on this
+//! fingerprint, so the splits leak straight through to codegen as duplicate
+//! `arrange_by_key`s — extra sorted indices rebuilt every recursive round.
+//!
+//! Canonicalizing the fingerprint fixes the sharing entirely in the plan: the
+//! existing dedup passes then collapse content-identical transformations, so the
+//! plan becomes a DAG with shared nodes and codegen lowers it 1:1 (no
+//! codegen-side sharing logic). This module produces the canonical operands the
+//! fingerprint is hashed from.
+//!
+//! # Canonicalization
+//!
+//! Within one transformation, every distinct atom (`AtomSignature`) is renamed
+//! to a canonical index by first-appearance order over a fixed traversal.
+//! `argument_id` (which column) is preserved; `rhs_id` and `is_positive` are
+//! normalized away. Conjunctive predicate lists are additionally order-
+//! normalized (sorted by element fingerprint) so two rules that state the same
+//! filters in a different textual order still share.
+//!
+//! # Soundness
+//!
+//! Interning keys on the *full* `AtomSignature` (polarity + position), so
+//! distinct atoms always receive distinct canonical indices — the renaming is
+//! injective and never merges genuinely different transformations. Input
+//! fingerprints are already canonical (a leaf's is the relation-name hash; a
+//! derived input's is the canonical fp of its producer), so canonicalizing each
+//! transformation's own operands composes bottom-up. Two transformations
+//! collide only when they have identical input fps, identical column structure
+//! (modulo position), and the identical predicate set — i.e. they provably
+//! produce the same collection.
+//!
+//! Polarity is content-irrelevant because an arrangement holds the projected
+//! relation data; the positive/negative distinction lives in the *consumer*
+//! (a semijoin vs an antijoin), both of which want the identical keyed trace.
+//!
+//! The canonical operands are used *only* for the fingerprint hash; the stored
+//! layouts/predicates keep their original form, so codegen is unaffected.
+
+use crate::catalog::{
+    ArithmeticPos, AtomArgumentSignature, AtomSignature, ComparisonExprPos, FactorPos,
+    FnCallPredicatePos, JoinPredicates, KvPredicates,
+};
+use crate::common::compute_fp;
+use std::collections::HashMap;
+use std::hash::Hash;
+
+use super::KeyValueLayout;
+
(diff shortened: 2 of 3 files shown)


hdz284:
⚠️ **Correctness regression on self-joins — please hold merge.**

The content-canonical fingerprint over-shares when the same relation appears at
multiple body positions: it collapses the two occurrences in a self-join to one
fingerprint, so the existing dedup eliminates one input arrangement and codegen
emits `R_arr.join_core(R_arr, …)` (joining one arrangement with itself).

Repro (clean builds, batik-independent — uses the in-repo fixtures):

```
# on main-next
bash tests/fixtures/run_compiler.sh join_self syntax_hybrid   # ✓ both pass

# on this branch (perf/share-arrangements-planner)
bash tests/fixtures/run_compiler.sh join_self syntax_hybrid   # ✗ both fail (output mismatch)
```

For `datalog-batch/join_self` (`TwoHop(a,c) :- Edge(a,b), Edge(b,c).`) the
generated dataflow becomes `t_edge_arr.join_core(t_edge_arr, …)`, yielding
self-loops (`1 1`, `2 2`, …) and `Triangle` = 1 row instead of 6.
`datalog-batch/syntax_hybrid` (`arc(a,c) :- arc(a,b), arc(b,c).`) fails the same
way. The 11 extra shares this captures over a lineage/rendered-code key include
these unsound self-join merges.

#142 is an alternative that shares at the **arrangement** level via plan-time
aliasing (each transformation keeps its own collection identity), so self-join
operands stay distinct — it keeps all 111+111 fixtures green and captures the
sound subset (1178 vs 1167 here). Either fix the canonicalization to not merge
distinct self-join operands, or take the #142 aliasing approach.

hdz284:
Consolidated into #142. Recapping the self-join issue flagged above: making the fingerprint content-canonical collapses plan *nodes*, so a self-join's two distinct keyings of one relation fuse (verified failing on `join_self` / `syntax_hybrid`). #142 keeps nodes distinct and shares at the *arrangement* level instead — same planner-only philosophy, self-join-sound, all fixtures green. Suggest closing this in favor of #142.

hdz284:
Closing in favor of **#142** (plan-time arrangement *aliasing*).

This PR's mechanism — making the collection **fingerprint** content-canonical so the existing dedup passes merge duplicates — is **unsound on self-joins**. Merging the two occurrences of a relation in `R(x,y), R(y,z)` to one collection fingerprint makes both join operands resolve to the same arrangement.

Verified on the `datalog-batch/join_self` fixture: `TwoHop`/`Triangle` come out wrong (spurious self-loops, missing tuples) under this branch, while `main-next` and #142 are correct. (The earlier DOOP/context-insensitive `.printsize` count checks passed only because those programs don't hit the exact collapse — the fixture suite is the real gate.)

#142 keeps each transformation's collection identity distinct and shares only the *arrangement* (a rule-independent `arrangement_key`, decided in the planner, rendered as a `clone()` by codegen), which is sound on self-joins and yields the same wins (DOOP 204→192, context-insensitive peak RSS −19%, ~−14% wall).

PR #142, opened 2026-06-11
perf(planner): cross-rule arrangement sharing, content-canonical & self-join-sound (consolidates #139/#141)

hdz284:
## Summary

A single, clean **planner-only** implementation of cross-rule arrangement
sharing — consolidating the three in-flight attempts (#139 codegen-CSE, #141
content-canonical fingerprint, and this branch's original aliasing) into one
approach that is **content-optimal-ish, self-join-sound, and keeps all sharing
logic in the query plan**. Codegen merely renders the planner's decision.

FlowLog rebuilds the same differential-dataflow **arrangement** many times: a
relation projected and keyed identically across rules is sorted and indexed once
per rule. On wide program-analysis workloads (DOOP / context-insensitive
points-to) a few relations are joined across hundreds of rules, so every
redundant arrangement is an extra sorted index rebuilt on every recursive round.

## Root cause

Arrangements are keyed by their collection **fingerprint**, which embeds the
rule-local atom signature (`rhs_id`) of the recipe that built them. Two rules
that build a byte-identical sorted trace via differently-shaped bodies get
distinct fingerprints, so the planner's existing dedup passes
(`deduplicate_transformation_infos`, `prune_cross_stratum_duplicates`) never
collapse them.

## Why aliasing, not fingerprint-merging (the soundness story)

The tempting "pure" fix is to make the **fingerprint itself** content-canonical
so the *existing* dedup collapses the duplicates with zero codegen changes —
this is #141. **It is unsound for self-joins.** Collapsing two content-identical
nodes merges plan *nodes*, but a self-join `R(x,y), R(y,z)` needs two *distinct*
keyings of `R`; once merged, the join references one arrangement for both sides.

Verified on a minimal `TwoHop(a,c) :- Edge(a,b), Edge(b,c)`:

| | join's two inputs |
| --- | --- |
| `main-next` (correct) | `L=proj_keycol1`, `R=proj_keycol0` — **distinct** |
| #141 | `L=R=proj_keycol1` — **fused** → self-loops (`1 1`), `Triangle`=1 row not 6 |

The two in-repo fixtures `datalog-batch/join_self` and `datalog-batch/syntax_hybrid`
fail on #141 and pass here.

So this PR keeps **every plan node distinct** (joins still reference the correct
two inputs) and shares only the physical **arrangement** when two distinct nodes
provably build the same sorted trace.

## Design — all in `planner/`

- **`Transformation::arrangement_key(input_keys)`** — a rule-independent identity
  of the arrangement a node builds: the *canonical keys of its inputs*, the
  *positional* flow content key, and the arrange kind (`arrange_by_self` vs
  `arrange_by_key`). It is rule-independent because the flow references key/value
  *positions*, never `rhs_id`.
- **`TransformationFlow::arrangement_content_key()`** — hashes the flow with the
  output key/value position lists **ordered** (they define the tuple layout, and
  are the self-join soundness boundary) but the conjunctive filters
  (`compares` / `fn_call_preds`) as an **order-independent multiset**. This is
  the extra sharing #139 noted a structural key over-splits on.
- **`StratumPlanner::build_arrangement_keys()`** — propagates keys in topological
  order from *canonical input keys*, so sharing is **transitive** (a join over
  already-shared inputs shares in turn).
- **`ProgramPlanner::mark_shared_arrangements()`** — a sibling of
  `prune_cross_stratum_duplicates`, keyed on `arrangement_key`. Marks each
  arrangement **canonical** or **alias-of-X**, scope-aware to match the existing
  **factoring** framework: program-wide for non-recursive (outer) work,
  per-stratum for recursive (iterative) work. Cross-scope reuse stays the
  existing `.enter()` mechanism.
- **codegen** (`flow/transformation.rs`) contains no sharing logic — an alias
  emits `let out = canonical.clone();` and points its fingerprint at the
  canonical's already-built trace, instead of rebuilding the projection/join +
  sorted arrangement.

## Results (batik, `--str-intern`, interleaved A/B vs `main-next`, `-w 16`, 3 reps)

| Program | metric | `main-next` | this PR | Δ |
| --- | --- | ---: | ---: | ---: |
| context-insensitive (592 rules) | arrangements | 1290 | 1175 | **−8.9%** |
| context-insensitive | solve | 58.9 s | 54.6 s | **−7.4%** |
| context-insensitive | userCPU | 933 s | 865 s | **−7.3%** |
| context-insensitive | peak RSS | 11806 MiB | 9712 MiB | **−17.7% (−2.05 GiB)** |
| DOOP (context-sensitive) | arrangements | 204 | 192 | −5.9% |

DOOP perf is flat — its shareable arrangements are small static EDB projections.
The payoff scales with program size.

## Correctness

- **111/111** compiler + **111/111** lib fixtures, incl. the self-join fixtures
  `join_self` / `syntax_hybrid`; **218** unit tests (new tests cover
  rule-independence, the key/column soundness boundary, and filter-order
  independence); `clippy` + `fmt` clean.
- **DOOP**: relation sizes identical (`VarPointsTo`=37,237,353, …).
- **context-insensitive**: all **21** output relations have identical row counts;
  the deterministic relations are **byte-identical**, and the points-to relations
  (`VarPointsTo`=28,249,074, …) differ only by the pre-existing benign `ord()`
  heap-representative nondeterminism the baseline itself exhibits run-to-run.

## Relation to #139 / #141

- **#139** (codegen CSE) — same goal; this keeps the logic in the planner instead
  of hashing rendered codegen tokens.
- **#141** (content-canonical fingerprint) — same goal; but #141 collapses plan
  nodes and is **unsound on self-joins** (see above). This captures the sound
  subset (1175 vs #141's partly-unsound 1167) with all fixtures green.

Recommend closing #139 and #141 in favor of this. A full Soufflé oracle run
before merge is advisable, per the repo's correctness gate.

### DIFF
--- a/crates/flowlog-build/src/codegen/flow/transformation.rs
+++ b/crates/flowlog-build/src/codegen/flow/transformation.rs
@@ -34,6 +34,15 @@ impl CodeGen {
         stratum: &StratumPlanner,
         profiler: &mut Option<Profiler>,
     ) -> Result<TokenStream, CodegenError> {
+        // Arrangement sharing is decided in the planner. If this output's
+        // arrangement was marked a duplicate of an earlier one in this scope,
+        // resolve the canonical collection ident to clone instead of
+        // rebuilding the projection + sorted trace (see
+        // `ProgramPlanner::mark_shared_arrangements`).
+        let arr_alias = stratum
+            .arrangement_alias_of(transformation.output().fingerprint())
+            .map(|canon_fp| find_local_ident(local_fp_to_ident, canon_fp));
+
         // `atom_fps` decides *which* inputs are named atoms; the label
         // text comes from `display_name` (the user's spelling).
         let atom_fps = stratum.atom_fps();
@@ -213,18 +222,16 @@ impl CodeGen {
                         .flat_map(|#row_pat: #row_ty| { #flat_map_body });
                 };
 
-                // Arrangement registration
-                let arrange_stmt = self.register_arrangement(
+                // Arrangement registration (shares a byte-identical arrangement
+                // already built in this scope, if any).
+                Ok(self.register_arrangement(
                     arranged_map,
+                    arr_alias.as_ref(),
                     output.fingerprint(),
                     &out,
+                    transformation,
                     output.is_k_only(),
-                );
-
-                Ok(quote! {
-                    #transformation
-                    #arrange_stmt
-                })
+                ))
             }
 
             // KV -> Row
@@ -362,17 +369,14 @@ impl CodeGen {
                 };
 
                 // Arrangement registration
-                let arrange_stmt = self.register_arrangement(
+                Ok(self.register_arrangement(
                     arranged_map,
+                    arr_alias.as_ref(),
                     output.fingerprint(),
                     &out,
+                    transformation,
                     output.is_k_only(),
-                );
-
-                Ok(quote! {
-                    #transformation
-                    #arrange_stmt
-                })
+                ))
             }
 
             // Join: Key-value ⋈ Key-value -> Row
@@ -501,17 +505,14 @@ impl CodeGen {
                         .join_core(#r.clone(), |#jn_k, #jn_lv, #jn_rv| { #join_body });
                 };
 
-                let arrange_stmt = self.register_arrangement(
+                Ok(self.register_arrangement(
                     arranged_map,
+                    arr_alias.as_ref(),
                     output.fingerprint(),
                     &out,
+                    transformation,
                     output.is_k_only(),
-                );
-
-                Ok(quote! {
-                    #transformation
-                    #arrange_stmt
-                })
+                ))
             }
 
             // Antijoin: Key-value ¬ Key-only to Row
@@ -659,15 +660,14 @@ impl CodeGen {
 
                 let arrange_stmt = self.register_arrangement(
                     arranged_map,
+                    arr_alias.as_ref(),
                     output.fingerprint(),
                     &out,
+                    transformation,
                     output.is_k_only(),
                 );
 
-                Ok(quote! {
-                    #transformation
-                    #arrange_stmt
-                })
+                Ok(arrange_stmt)
             }
         }
     }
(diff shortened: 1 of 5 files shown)


hdz284:
## Independent verification + comprehensive benchmark

I reviewed this branch against the other sharing/factoring PRs (#139 codegen-CSE, #141/#143 fingerprint-merge) and re-benchmarked it across the program suite and multiple datasets. **This branch is the right consolidation**: it takes the merits of all of them — content-based sharing (#139), order-independent conjunctive predicates (#141's completeness idea, now in `arrangement_content_key`), and a planner-phase decision — while being the only one **sound on self-joins** (it aliases the *arrangement* and keeps each collection's identity distinct, instead of merging collection fingerprints, which collapses the two operands of `R(x,y), R(y,z)`).

### Correctness
- **111/111 fixtures** + 13 unit-test groups pass; clippy clean.
- Output identical (base vs this branch) on **every (program × dataset) cell below** — including `cspa`, whose recursive self-join `ValueFlow(x,y) :- ValueFlow(x,z), ValueFlow(z,y)` is exactly the pattern that makes the fingerprint-merge approach (#141) miscompile. DOOP `.printsize` 26/26 and context-insensitive per-relation counts match.

### Arrangements eliminated — all-programs sweep (compile-time, dataset-independent)
```
Arrangements eliminated (compile-time; bar = % saved)
  sg                ██████████████████████████████ 25%  (4->3)
  cspa              ████████████████████████ 20%  (10->8)
  context-insens.   ██████████▍ 8.6% (1289->1178)
  crdt              █████████▎ 7.7% (26->24)
  galen             ████████ 6.7% (15->14)
  doop              ███████▏ 5.9% (204->192)
  cvc5 / z3         ████▍ 3.6% (83->80)
```
No-op (0 change, **0 regression**) on tc / reach / cc / sssp / bipartite / dyck / andersen / csda / polonius_str. Suite total **1800 → 1665 (−7.5%)**.

### Peak RSS — flagship, holds across worker counts
```
Peak RSS — context-insensitive / batik (lower is better)
  base   -w8  █████████████████████████████▉ 11652 MB
  shared -w8  ███████████████████████▋  9216 MB  -20.9%
  base   -w16 ██████████████████████████████ 11710 MB
  shared -w16 ████████████████████████▊  9652 MB  -17.6%
```

### Peak RSS reduction across programs & datasets (-w8)
```
Peak RSS reduction by program/dataset (bar = % saved, -w8)
  context-insens. / batik  ██████████████████████████████ -20.9%
  context-insens. / avrora ██████████████████▊ -13.1%
  cspa / linux             ███████████▍ -7.9%
  cspa / postgresql        ███▎ -2.3%
  doop / batik             ██▉ -2.0%
  cspa / httpd             ██▏ -1.5%
  galen                    █ -0.7%
```

### Full runtime table (all `OK` = output identical to baseline)
| program / dataset | workers | wall Δ | peak RSS Δ | correct |
| --- | --- | ---: | ---: | --- |
| context-insensitive / batik | 8 | −9.1% | **−20.9%** (11652→9216 MB) | OK |
| context-insensitive / batik | 16 | −10.2% | **−17.6%** (11710→9652 MB) | OK |
| context-insensitive / avrora | 8 | −3.1% | **−13.1%** (2728→2370 MB) | OK |
| cspa / linux (self-join) | 8 | −3.1% | −7.9% (3223→2969 MB) | OK |
| cspa / postgresql (self-join) | 8 | −0.8% | −2.3% | OK |
| cspa / httpd (self-join) | 8 | +0.6% | −1.5% | OK |
| doop / batik | 8 | ~flat | −2.0% (4240→4154 MB) | OK |
| doop / fop | 8 | ~flat | −0.9% | OK |
| doop / avrora | 8 | ~flat | −1.7% | OK |
| doop / luindex | 8 | ~flat | −1.1% | OK |
| galen | 8 | ~flat | −0.7% | OK |
| crdt | 8 | ~flat | noise (±10% on a 180 MB / 1.5 s run) | OK |

### Takeaways
- The win **scales with program size and arrangement reuse**: context-insensitive (592 rules, 111 shared arrangements) gets **−18…21% peak RSS** and **−9…10% wall**, and it generalizes across datasets (batik, avrora).
- DOOP-135's 12 shared arrangements are small static EDB projections → low-single-digit RSS, wall within noise. Programs with no cross-rule reuse are untouched (no regression).
- `cspa` confirms self-join **soundness on real recursive data** at scale.

Recommendation: this is the one clean planner-phase branch; **#139** (codegen CSE) is superseded by it and can be closed.

HarukiMoriarty:
Agree with the basic idea, but I am more prefer minimizing the coding changes, and should be only modified per rule planning logic, the cross-rule and cross-stratum logic should take the free ride.
For example: one way I am thinking is modify the `AtomArgument` so `rhs_id` is participate in the planning but has no influence about the `fingerprints`. 
Another way you can refer to this branch `nemo/perf-canonical-fp`, where I put rehash logic immediately after the `post` phase. Though I still do not think this branch do an elegant approach.

hdz284:
Superseded by #144 (`perf/west4-arrangement-sharing`), which is the same planner-only arrangement-aliasing approach plus a provable join/antijoin-soundness guard (operator tag + `semijoin_antijoin_share` fixture) and order-independent conjunctive-filter/equality-constraint content keys. Closing in favor of #144; same content-optimal arrangement counts (context-insensitive 1289→1177, doop 204→192) and the same correctness gates.

PR #144, opened 2026-06-11
[west 4] perf(planner): cross-rule arrangement sharing

hdz284:
Cross-rule **arrangement sharing**, resolved entirely in the planner. Codegen just renders the decision.

## What & why

FlowLog re-sorts the same differential-dataflow arrangement once per rule. A relation projected and keyed identically across differently-shaped rules gets *distinct* lineage fingerprints (they embed the rule-local atom position `rhs_id`), so the existing dedup passes never collapse them. On wide program-analysis workloads a few relations are joined across hundreds of rules — each redundant arrangement is an extra sorted index rebuilt on every recursive round.

## How (planner-only)

- **`Transformation::arrangement_key`** — a rule-independent identity of the arrangement a node builds: canonical keys of its inputs + the *positional* flow content + the operator tag.
- **`StratumPlanner::build_arrangement_keys`** — propagates keys in topological order, so sharing is **transitive** (a join over already-shared inputs shares in turn).
- **`ProgramPlanner::mark_shared_arrangements`** — a sibling of `prune_cross_stratum_duplicates`, scope-aware (program-wide for non-recursive, per-stratum for recursive). Marks each arrangement **canonical** or **alias-of-X**.
- **codegen** — an alias emits `let out = canonical.clone();` instead of rebuilding the projection + sorted arrangement.

**Self-join sound.** Every plan node stays distinct (a self-join keeps its two distinct keyings); only the *physical* arrangement is shared. The operator tag also keeps a semijoin and an antijoin — which both lower to a `JnToKV` flow — from ever aliasing to complementary data.

## Results

| program | arrangements | peak RSS (−w32, 20 DaCapo) |
|---|---|---|
| `context-insensitive` (592 rules) | 1289 → **1178** (−8.6%) | mean **−10%**, up to **−16.6%** (jython **−26 GiB**) |
| `doop` | 204 → **192** (−5.9%) | flat (its shareable arrangements are tiny static EDB projections) |

Solve time is neutral-to-positive, scaling with problem size (e.g. batik −6.5% @ −w16). Sharing generalizes across the suite: `sg` 4→3, `cspa` 10→8, `galen` 15→14; `tc`/`reach`/`cc`/`sssp`/… untouched (0 regression).

## Correctness

- 218 unit tests · **112 compiler + 112 lib fixtures** (incl. the self-join fixtures `join_self`, `syntax_hybrid`, and a new `semijoin_antijoin_share`) · `clippy` + `fmt` clean.
- Per-relation row counts **identical to `main-next`** across all 20 DaCapo apps for both `doop` and `context-insensitive`; the recursive self-join `cspa` is byte-identical. (The points-to relations differ only by the pre-existing benign `ord()` heap-representative non-determinism the baseline itself shows run-to-run.)

## Why not make the fingerprint content-canonical?

The tempting "free ride" — canonicalize the collection fingerprint so the *existing* dedup merges duplicates with no new code — is **unsound on self-joins**. Merging two content-identical *nodes* fuses a self-join's two keyings: verified on `join_self` (`TwoHop(a,c) :- Edge(a,b), Edge(b,c)`), where it collapses the col-0 and col-1 keyings of `Edge` into one and emits `edge_arr.join_core(edge_arr, …)` → wrong results. This PR shares at the **arrangement** level and keeps every node distinct.

Supersedes #139 / #141 / #142 / #143.


hdz284:
## Follow-up: equality constraints now order-independent too (1178 → 1167)

Reflecting on the content key, I found one conjunctive filter was still being
split by recording order. `TransformationFlow::arrangement_content_key` hashed
`compares` and `fn_call_preds` as order-independent multisets, but hashed the
equality **`constraints`** order-*sensitively* (`compute_fp(constraints)`).
Equality constraints are conjunctive too — `x = 5, y = z` keeps exactly the same
rows as `y = z, x = 5` — so two rules that recorded the same constraints in a
different order were building separate arrangements.

Fix (`bd27458`): added `Constraints::content_key()` (each kind hashed as a
sorted multiset) and used it, finishing the order-independence story for **all**
conjunctive filters.

### ⚠️ On the count: 1167 here is the *sound* 1167, not #141's
This lands on the same arrangement count as the **unsound** #141, but via a
different mechanism, and it is sound:
- We still alias only when the whole `arrangement_key` matches, and that key
  hashes the **key/value layout positions in order**. A self-join's two
  differently-keyed projections therefore keep distinct keys and can **never**
  fuse — the #141 failure mode is structurally impossible here. Only the
  *conjunction order* of equality filters is relaxed, which does not change
  arrangement content.
- The 11 newly-aliased arrangements are all plain `let t_x = t_y.clone();` over
  content-identical traces; none is a join operand collapse.

### Verification
- **220 unit tests** (incl. 2 new `Constraints::content_key` order/▵ tests) +
  **all 112 compiler fixtures** pass — `join_self`, `syntax_hybrid`,
  `semijoin_antijoin_share`, `recursive_tc_delta` included. clippy + fmt clean.
- **Real A/B on context-insensitive (`--str-intern`)**, base = current west4
  (1178) vs head (1167), `tomcat` + `batik`:
  - per-relation **row counts identical**;
  - the only byte-differing relations are exactly the set the **base shows
    against itself run-to-run** (ord() heap-representative nondeterminism:
    `VarPointsTo`, `*PointsTo`, `CallGraphEdge`, `Reachable`, …); deterministic
    relations byte-identical; sorted-set divergence base↔base 9.2% ≈ base↔head
    11.0% (same equivalence class).
- **Perf neutral** (the −0.9% arrangement delta is below the RSS noise floor):

  | app | metric | base (1178) | head (1167) |
  |---|---|---|---|
  | batik −w16 | solve (median) | 53.3 s | 53.7 s |
  | batik −w16 | peak RSS (median) | 9994 MB | 9971 MB |
  | tomcat −w16 | solve (median) | 13.40 s | 13.45 s |
  | tomcat −w16 | peak RSS (median) | 2889 MB | 2893 MB |

Net: a cleanliness/consistency fix that also captures 11 more **sound** shares
on context-insensitive, with no regression. Other programs (sg/cspa/galen/doop/
crdt/cvc5/z3) unchanged.

hdz284:
## Tidy-up pass + re-benchmark (force-pushed `bd27458` → `0cabd38`)

Cleaned the previous follow-up so the three conjunctive-filter kinds are handled
**consistently** instead of with a duplicated inline pattern:

- Factored the "fingerprint each item → sort → combine" multiset hash into one
  `common::compute_fp_unordered` helper (sibling of `compute_fp`).
- `arrangement_content_key` now routes `compares` + `fn_call_preds` through it,
  and `Constraints::content_key` routes both equality kinds through it — one
  definition, three call sites, no inline duplication.
- Trimmed the doc comments.

Behavior-identical to the prior commit: **context-insensitive still 1167**
arrangements (sg/cspa/galen/doop/crdt/cvc5/z3 unchanged), 220 unit tests + all
112 compiler fixtures green, clippy + fmt clean.

### Re-benchmarked correctness (context-insensitive, `--str-intern`, tomcat)
Per-relation **row counts identical** to base. Byte differences are confined to
the `ord()` heap-representative / parallel output-ordering set the **base shows
against itself run-to-run** (control: base-vs-base differs in the same
relations); deterministic relations are byte-identical, and small relations are
identical once sorted.

### Re-benchmarked performance — neutral (solve delta is layout noise)
The honest catch: comparing two separately-compiled ~165 MB binaries, solve time
carries a layout-noise floor. I verified it by building **two head binaries from
byte-identical source** and round-robining base/head/head2 on tomcat `-w16`:

| binary | solve median | vs base |
|---|---|---|
| base (1178) | 13.48 s | — |
| head (1167) | 13.72 s | +1.8% |
| head2 (1167, identical source to head) | 13.56 s | +0.5% |

**head vs head2 (identical source) = +1.2%** — i.e. the base/head/head2 solve
times all sit inside the layout-noise envelope, and RSS is flat across all three
(2890–2902 MB). The −0.9% arrangement reduction is real but below the wall-clock
measurement floor at this scale; removing arrangements cannot add solve work.

Net unchanged: a cleanliness/consistency fix that also captures 11 more **sound**
shares on context-insensitive, no regression.

hdz284:
Pushed a correctness fix (8adba9f) found during a review pass.

**Bug (extended mode only):** `mark_shared_arrangements` deduplicated non-recursive arrangements **program-wide across strata** using only the value-insensitive `arrangement_key`, with no check that the canonical's value is still current. In extended mode a `fixpoint`/`loop` can rewrite an IDB between two strata, so a later read of the *grown* IDB was aliased to an earlier read of the *stale* (seed) arrangement → silent wrong results. `datalog-batch` never rewrites an IDB across strata, so DOOP/context-insensitive were unaffected; the existing `loop_*` fixtures didn't read an IDB in a plain segment both before and after the loop that updates it, so they missed it.

Minimal repro (`--mode extend-batch`, `E = 1→2→3→4`):
```
Z(a,b):-E(a,b).  R1(a,c):-Z(a,b),E(b,c).  fixpoint{ Z(a,c):-Z(a,b),E(b,c). }  R2(a,c):-Z(a,b),E(b,c).
```
R2 must be {1→3,1→4,2→4} (grown Z) but came out {1→3,2→4} (= R1, stale seed).

**Fix:** gate cross-stratum non-recursive aliasing with the same transitive intervening-IDB-update check as the sibling `prune_cross_stratum_duplicates` (reuse a canonical at stratum `idx` only if no IDB it transitively depends on was written at a stratum `k` with `prev < k <= idx`; else it becomes the new canonical). **No-op in batch** — context-insensitive arrangement count stays 1167 and batch output is byte-identical. Adds regression fixture `extend-batch/arrangement_share_fixpoint_idb` (self-alias + cross-alias paths; fails before, passes after).

Validation: 220 unit tests, clippy + fmt clean, all 112 e2e fixtures pass, both repros correct.

hdz284:
Superseded by #145, which lands the same sharing **in the planner only** (content-canonical fingerprint rehash, the free-ride approach @HarukiMoriarty asked for in #142) — codegen is untouched, ~⅓ the code, and it captures one more sound share (1167 vs this branch's 1178). Verified self-join-sound (join_self / syntax_hybrid / swapped_columns / semijoin_antijoin_share) and benchmarked on context-insensitive (peak RSS −10…20%, solve neutral-to-faster, per-relation row counts identical to main-next). Closing this codegen-aliasing branch in favor of #145.

PR #145, opened 2026-06-11
perf(planner): cross-rule arrangement sharing via content-canonical fingerprints

hdz284:
Cross-rule **arrangement sharing**, resolved entirely in the planner — the free-ride
approach @HarukiMoriarty asked for in #142. Codegen is untouched.

This supersedes the codegen-aliasing branch in #144: same wins, ~⅓ the code, no codegen change.

## Why

FlowLog re-sorts the same differential-dataflow arrangement once per rule. A relation
projected and keyed identically across differently-shaped rules gets *distinct* lineage
fingerprints (they embed the rule-local atom position `rhs_id`), so the existing dedup
passes never collapse them. On wide program-analysis workloads a few relations are joined
across hundreds of rules, and each redundant arrangement is an extra sorted index rebuilt
every recursive round.

## How (planner-only, "free ride")

After `post` (the last lineage-fingerprint mutation), `RulePlanner::canonicalize_fingerprints`
rehashes every fingerprint into a **content-based canonical value**, so fingerprint equality
now means *content* equality. The within-stratum dedup, the cross-stratum prune, and codegen's
fp-keyed arrangement reuse then collapse the duplicates **with no further changes**.

- `TransformationInfo::flow()` — single source of truth for the data-movement flow, shared by
  the `Transformation` constructors and the canonical key.
- `TransformationInfo::canonical_fp` — operator kind + inputs (resolved to canonical) + the
  flow's positional `content_key`.
- `TransformationFlow::content_key` — output key/value layout hashed **in order**; conjunctive
  filters (equalities, compares, fn-call preds) hashed **order-independently**.
- A debug reproducibility guard asserts every rehashed fp is the canonical fp of its own final
  fields — the "equal fingerprint ⇒ equal content" invariant the dedup relies on.

## Self-join sound

The canonical key hashes the key/value layout **positions in order**, so a self-join's two
differently-keyed projections (`TwoHop(a,c) :- Edge(a,b), Edge(b,c)`) keep distinct keys and
can never fuse — the failure mode of the naive "content-canonical collection fingerprint" in
#141. Only the conjunction *order* of commutative filters is relaxed, which does not change
arrangement content. Pinned by fixtures `join_self`, `syntax_hybrid`, `swapped_columns`, and
`semijoin_antijoin_share`.

## Results (`context-insensitive`, 592 rules, `--str-intern`)

Arrangements: **1287 → 1167 (−9.3%)** — 1287 main-next → 1178 content-canonical →
1167 once equality constraints are treated order-independently too (+11 sound `by_key` shares).

Fresh A/B, `ci_base` (main-next) vs `ci_head` (this PR), `−w16`, 21 output relations each:

| dataset | solve Δ | peak RSS Δ | row counts |
|---|---:|---:|---|
| batik  | −9.1 % | **−18.5 %** (11766→9585 MB) | identical |
| h2     | −9.3 % | −14.0 % (12436→10693 MB) | identical |
| fop    | −5.7 % | −15.9 % (8314→6991 MB) | identical |
| xalan  | −1.3 % | −10.9 % | identical |
| avrora | −1.8 % | −12.2 % | identical |
| tomcat | −5.1 % | −13.9 % | identical |

Solve is neutral-to-faster and scales with problem size; peak RSS drops **~10–20 %**. `doop`
204 → 192 (its shareable arrangements are tiny static EDB projections → RSS flat). Programs with
no cross-rule reuse are untouched (0 regression).

## Correctness

Per-relation **row counts identical to main-next** on every dataset. Byte-level (sorted): 19/21
relations identical; the 2 that differ (`VarPointsTo`, `InstanceFieldPointsTo`) differ in exactly
the same way the **base differs against itself run-to-run** — the pre-existing benign `ord()`
heap-representative nondeterminism (verified with a base-vs-base control). Deterministic relations
are byte-identical.

**218 unit tests · 113 compiler + 112 lib fixtures · clippy + fmt clean.**

---
Built on @hdz284's clean `flow()`-refactored canonical base (58457d2); this PR adds the
order-independent equality constraints + the reproducibility soundness guard on top.


hdz284:
Confirmed the peak-RSS win holds across worker counts (flagship `batik`, fresh A/B, all 21 relations row-count-identical):

| workers | solve Δ | peak RSS Δ |
|---|---:|---:|
| −w8  | −12.9 % | −19.8 % (11439→9172 MB) |
| −w16 | −9.1 %  | −18.5 % (11766→9585 MB) |

So removing the redundant per-rule arrangements cuts both wall-clock and memory at both worker counts; the effect doesn't wash out as parallelism rises.

hdz284:
Independently re-vetted + benchmarked this branch (1d5f5a4):

**Correctness** — 218 unit tests, clippy + fmt clean, **all 114 e2e fixtures pass** (incl join_self / syntax_hybrid / swapped_columns / semijoin_antijoin_share). Crucially, this free-ride approach is **immune by construction to the cross-stratum stale-IDB hazard** that the codegen-aliasing branch (#144) hit: there's no separate sharing pass, so the existing `prune_cross_stratum_duplicates` (with its intervening-IDB-update check) governs cross-stratum reuse. Verified on a `fixpoint` repro where a relation is re-keyed before and after the loop that grows it — the post-loop read correctly sees the grown IDB (the aliasing branch returned the stale seed before its fix). Added a regression fixture `extend-batch/arrangement_share_fixpoint_idb` (d3d9d81) to pin it (self-alias + cross-alias paths).

**Arrangements** — context-insensitive 1287 → **1167** confirmed; doop 204 → 192.

**Perf A/B vs main-next** (`-w16`, 3-rep medians):
| dataset | solve Δ | peak RSS Δ | correctness |
|---|---:|---:|---|
| batik | −10.0% | **−17.8%** (11.57→9.51 GB) | resid=0 ALL_BENIGN |
| h2 | −9.8% | −17.4% (12.61→10.41 GB) | resid=0 ALL_BENIGN |

Matches your reported numbers. LGTM — smaller, codegen-free, same win, and architecturally the safer place for the cross-stratum check (it's reused, not reimplemented).

PR #152, opened 2026-06-12
test(planner): cover the load-bearing cross-stratum gate; add producer-split regressions + precise docs

hdz284:
## Background

This started from a correctness question about the cross-stratum **factoring**
optimization (the slide-deck framing *“once a stratum finishes, its relations
never change again — fixed”*), with this counterexample:

```prolog
IDB1(x) :- EDB1(x).
IDB2(x) :- EDB2(x).
IDB1(x) :- IDB2(x).
```

> Isn’t `IDB1` *not* fixed after the first stratum, since the third rule adds to it later?

## Finding

**The observation is correct, and it is not a bug.** FlowLog stratifies at the
**rule level**, so a relation’s *producer* rules can land in different strata.
Confirmed by the stratifier:

```
#1 [non-recursive] [0, 1]   idb1(x) :- edb1(x).   idb2(x) :- edb2(x).
#2 [non-recursive] [2]      idb1(x) :- idb2(x).
```

`IDB1` heads **both** stratum 0 and stratum 1, so it is genuinely *not* “fixed”
after stratum 0. It’s sound anyway because of the **dual property**: a *consumer*
of a relation depends on **every** producer of it and is therefore stratified
strictly after all of them — so whenever a relation is *read* (and thus arranged,
the unit factoring reuses) it is already complete. `prune_cross_stratum_duplicates`’
intervening-IDB-update gate is a **no-op in `datalog-batch`** and is needed only
in **extended** mode, where a `loop`/`fixpoint` can *rewrite* an IDB between two
reads across a barrier.

This PR is **docs + tests only — no engine behaviour change.**

---

## 1. Proven coverage gap: the gate was load-bearing but untested

The gate lives in `prune_cross_stratum_duplicates`:

```rust
Some(&prev) => t_deps.iter()
    .any(|idb| idb_writes[idb].iter().any(|&k| k > prev && k <= idx)),
```

**Experiment (clean `main-next`, gate deleted → `… => false`):**

| run | result |
|-----|--------|
| `cargo test -p flowlog-build` | **274 passed, 0 failed** |
| full `run_lib.sh` e2e suite | **111 passed, 0 failed** |

i.e. **nothing in the repo exercised the gate** — you can delete load-bearing
code and stay green.

This PR adds **`extend-batch/arrangement_share_fixpoint_idb`** (Z is re-keyed
before a `fixpoint` that grows it, and again after). With the gate deleted it
fails with silent wrong results; with the gate it passes:

```
✗ arrangement_share_fixpoint_idb — output mismatch   (gate deleted)
    R2: expected 3 rows, got 2   (missing 1→4)
    R3: expected 6 rows, got 3   (missing 1→2, 1→3, 2→3)
```

A/B confirmed on the current `main-next` engine **and** the prior branch.

## 2. Scenario coverage: producer rules that split across strata

I tried to find a “fixed-after-stratum” mutation that **only** these fixtures
catch. I couldn’t: a broad `leave`-set version is caught by existing recursive
fixtures (`recursive_tc`, …), and a narrow `prune`-relation-key version is caught
by existing union fixtures (`rule_union`, …) via build failures. So the
underlying non-recursive union/leave machinery is **already partly guarded** —
these are **regressions for the exact reported scenario**, not a second gap:

- **`stratifier::producer_rules_split_across_strata`** — asserts both the
  producer-split *and* the consumer-comes-later property (no existing test
  asserts either; e.g. `available_set_accumulates_leaves_across_strata` uses a
  single-producer relation).
- **4 `datalog-batch` fixtures** — `producer_split_{union, read_downstream,
  selfjoin, into_recursion}` cover the split being unioned, read downstream,
  self-joined, and entered into a recursive loop. Outputs **verified against
  Soufflé**; green in **compiler and lib** modes.

## 3. Docs

- Stratifier: `available` is *“readable, not necessarily complete”* (was
  “fully computed”), with the producer-split / consumer-comes-later rationale.
- `prune_cross_stratum_duplicates`: expanded the soundness note so the
  “fixed after a stratum” intuition isn’t repeated.

## Verification

- `cargo test -p flowlog-build`: **275 passed, 0 failed**.
- 5 new fixtures green in **both** compiler and lib modes.
- A/B proves the gate is load-bearing (extended) and a no-op (batch).
- No engine code changed.


hdz284:
Superseded by #157, which consolidates all the sharing/factoring findings (this PR's content included) into one clean, fully-tested PR.

PR #148, opened 2026-06-12
perf(planner): share arrangements across rules via content-canonical materialization

HarukiMoriarty:
## Bug

Planner dedup keys on info fingerprints, which embed the rule-local atom position (`rhs_id`). The same arrangement built by different rules gets different fingerprints, so it's rebuilt once per rule -- hundreds of redundant arrangements on DOOP-scale workloads.

## Fix

Materialize each rule's `TransformationInfos` into `Transformations` right after post, computing the output fingerprint from (variant tag, input fps, flow) -- the flow is positional, so `rhs_id` drops out. A per-rule map (info fp -> content fp) lets downstream inputs resolve to their producers. Stratum dedup then collapses identical operations across rules; everything downstream (idbâheads map, profiler, debug trees) reads the new fingerprints.

PR #153, opened 2026-06-12
refactor(planner): de-duplicate Transformation::from_info; document the sharing pipeline

hdz284:
## Summary

A readability/structure pass over the arrangement-**sharing** machinery (the
content-canonical materialization path, #148). **No behaviour change** — purely a
tidy-up of `Transformation::from_info` plus the documentation that was missing
around the whole sharing pipeline.

### `Transformation::from_info` cleanup

The three constructors (`kv_to_kv` / `join` / `antijoin`) each:
- rebuilt input/output `Collection` handles by hand (`Arc::new(Collection::new(..))` ×N), and
- recomputed the content-canonical output fingerprint,

and `from_info` matched the `info` variant **twice** — once to pick a `tag`
string, once to dispatch. This change:

- extracts a `collection(fp, name, layout)` helper used by every constructor;
- extracts `binary_collections(..)` for the input-pair + output shared by the
  binary constructors (`join`/`antijoin` differ only in predicates, the key-only
  assertion, and the variant names);
- moves each variant's `tag` next to the flow it fingerprints, so `from_info`
  dispatches **once**.

The `compute_fp` inputs — `(variant tag, resolved input fps, flow)` — are
**byte-for-byte unchanged**, so every content fingerprint (and therefore all
sharing/dedup/prune decisions and the generated dataflow) is identical.

### Docs

`planner/mod.rs` had no module doc; added an overview of how one
content-canonical fingerprint drives the four sharing passes:

> materialize (`from_info`) → within-stratum dedup → recursion factoring
> (`.enter()` static vs delta) → cross-stratum prune

…and a short note on `transformation`'s role.

## Verification

**Logic-preserving (generated-code parity).** Generated `main.rs` is identical —
after normalising codegen's *pre-existing* nondeterminism (random temp-var names;
commuted equality-filter conjunctions, which I confirmed flip run-to-run on the
**baseline** binary too) — across:

| program | shared arrangements | normalized diff vs baseline |
|---|---|---|
| `doop.dl` | 192 | = same-binary nondeterminism |
| `polonius_int.dl` | 56 | 0 |
| `polonius_str.dl` | 56 | 0 |
| context-insensitive `batik.dl` | 190 | = same-binary nondeterminism |

**Tests.** `cargo test -p flowlog-build`: 274 passed, 0 failed. Full lib fixture
suite: 111 passed, 0 failed.

**Benchmark (no regression).** `doop.dl` @ DaCapo `batik` (1.7 GB facts, `-w 16`):

| | run 1 | run 2 | peak RSS |
|---|---|---|---|
| baseline | 107.24 s | 107.60 s | 52.88 GB |
| refactored | 107.89 s | 106.19 s | 52.90 GB |

Within noise — consistent with the byte-identical generated code. (Polonius/ctx
perf is identical by the same parity argument; `polonius_*` on `clap` is a
many-minute string workload, so timing it adds noise without information.)

Scope is disjoint from #152 (that PR touches `stratifier/core.rs` +
`program_planner.rs` + fixtures; this one touches `transformation.rs` +
`planner/mod.rs`), so they don't conflict.

### DIFF
--- a/crates/flowlog-build/src/planner/mod.rs
+++ b/crates/flowlog-build/src/planner/mod.rs
@@ -1,3 +1,38 @@
+//! Query planning: lower stratified rules into a shared dataflow plan.
+//!
+//! A [`ProgramPlanner`] turns a stratified program into one [`StratumPlanner`]
+//! per stratum. Within a stratum, each rule is planned independently into a
+//! pipeline of [`Transformation`]s (prepare → SIP → core join → fuse → post),
+//! and then the per-rule pipelines are **shared** so identical work is emitted
+//! once.
+//!
+//! # Arrangement / transformation sharing
+//!
+//! Sharing is driven entirely by a **content-canonical fingerprint** computed
+//! when each rule's plan is materialized, then exploited by three passes that
+//! all key on that fingerprint:
+//!
+//! 1. **Materialize** ([`RulePlanner::materialize`] → [`Transformation::from_info`]):
+//!    rewrite each operation's *lineage* fingerprint (which embeds rule-local
+//!    atom positions and so never matches across rules) into a content fp =
+//!    `hash(variant tag, resolved input fps, flow)`. Identical operations from
+//!    different rules now carry the same fingerprint.
+//! 2. **Within-stratum dedup** ([`StratumPlanner`]): keep the first occurrence
+//!    of each content fingerprint, in topological order.
+//! 3. **Recursion factoring** ([`StratumPlanner`]): split a recursive stratum's
+//!    transformations into *static* (EDB / earlier-stratum inputs — built once
+//!    and `.enter()`-ed into the loop) and *dynamic* (IDB-dependent — rebuilt
+//!    each round), so loop-invariant arrangements are not rebuilt per round.
+//! 4. **Cross-stratum prune** ([`ProgramPlanner`]): drop a non-recursive
+//!    transformation whose content fingerprint was already emitted by an
+//!    earlier stratum, unless an intervening stratum rewrote an IDB it depends
+//!    on (the extended-mode soundness gate).
+//!
+//! Because every pass keys on the same content fingerprint, two rules that
+//! reach the same arrangement via different shapes still share it, and a
+//! relation read after all its producers is always read in its complete form.
+
+
 mod argument;
 mod arithmetic;
 mod collection;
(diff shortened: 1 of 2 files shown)


crates/flowlog-build/src/planner/mod.rs
```diff
@@ -1,3 +1,38 @@
...
+//!
+//! 1. **Materialize** ([`RulePlanner::materialize`] → [`Transformation::from_info`]):
+//!    rewrite each operation's *lineage* fingerprint (which embeds rule-local
+//!    atom positions and so never matches across rules) into a content fp =
+//!    `hash(variant tag, resolved input fps, flow)`. Identical operations from
+//!    different rules now carry the same fingerprint.
+//! 2. **Within-stratum dedup** ([`StratumPlanner`]): keep the first occurrence
+//!    of each content fingerprint, in topological order.
+//! 3. **Recursion factoring** ([`StratumPlanner`]): split a recursive stratum's
+//!    transformations into *static* (EDB / earlier-stratum inputs — built once
```
HarukiMoriarty, line 23: Should optimization docs should be locate here? I basically think for here should introduce what is planner doing and that should be enough. Tech details like sharing optimization, fusion optimization should be placed at other places?

crates/flowlog-build/src/planner/transformation.rs
```diff
@@ -218,6 +228,18 @@ fn resolve_fp(fp_map: &HashMap<u64, u64>, fp: u64) -> u64 {
...
+/// so input/output handles are built one consistent way.
+fn collection(fp: u64, name: &str, layout: &KeyValueLayout) -> Arc<Collection> {
+    Arc::new(Collection::new(
+        fp,
+        name.to_string(),
+        layout.key(),
+        layout.value(),
+    ))
+}
+
```
HarukiMoriarty, line 242: Helper function includes later `binary_collection` should be placed at some block together.

HarukiMoriarty (commented):
The comments cleaning should be in a more careful way, especially optimization tricks should not be put at the crate entry level, that will significant increase the complexity of understanding.

hdz284:
Closing in favour of #155, which consolidates the three overlapping `from_info` cleanups (#153, #154, #155) that were opened within minutes of each other.

#155 keeps the spirit of this PR's collection-factoring (as a single `collection_of` helper) and uses the `(tag, ctor)` single-match form that fully removes the tag↔variant drift hazard. Per @HarukiMoriarty's review here — *optimization-trick docs shouldn't sit at the crate entry level* — #155 deliberately omits the `planner/mod.rs` module-overview docs this PR added.

One factual note for the record: this PR's verification table lists *context-insensitive `batik.dl` → 190 arrangements*, but the flattened context-insensitive program emits ~1177–1178 `arrange_by_*` (doop.dl is the one at ~192). Worth double-checking that figure if any of this is revived.

PR #154, opened 2026-06-12
refactor(planner): co-locate variant tags + factor collection-building in Transformation::from_info

hdz284:
## What

A behavior-preserving cleanup of the content-canonical materialization added in #148
(`Transformation::from_info` + the `kv_to_kv` / `join` / `antijoin` constructors).

Two things were awkward in the merged version:

1. **Tag/variant sync hazard.** `from_info` computed each operator's variant tag in a
   `match` over the `TransformationInfo` variant, then dispatched in a *second* `match`
   over the same variant right below it. The two had to stay in lock-step — the code even
   comments "*tag mirrors the variant picked below*". Parallel switches over the same thing
   are exactly the kind of coupling that rots.
2. **Repeated collection boilerplate.** Each constructor open-coded the same 4-line
   `Arc::new(Collection::new(fp, name.to_string(), layout.key(), layout.value()))` for every
   input and output (6 copies total).

## Change

- Each variant tag now lives **inside the constructor that builds that variant**, next to
  the variant it seeds. `from_info` reduces to: resolve inputs → dispatch → record the
  lineage→content fp mapping.
- Collection construction is factored into a small `collection_of(fp, name, layout)` helper.
- Comments tightened to explain the lineage-fp → content-fp resolution and why the tag
  exists (so a row/kv shape difference can't collapse two otherwise-identical flows onto
  one arrangement).

Net **−25 lines**, no new types, clippy clean.

## No behavior change

The tag strings (`row_to_row`, `jn_to_kv`, …) and the `compute_fp` tuple shapes are
preserved byte-for-byte, so content fingerprints — and therefore cross-rule arrangement
sharing — are identical. Verified:

| check | result |
|---|---|
| `cargo test -p flowlog-build` | **218 passed** (incl. `equal_fingerprint_implies_equal_content`, rhs_id sharing, swapped-columns tripwire, dyck prune) |
| arrangement counts (generated `arrange_by_*`) | **polonius 56, context-insensitive 1177, doop 192** — unchanged vs `main-next` |
| ci@tomcat output | **21/21 relations identical to Souffle** (`VarPointsTo`=3,066,514) |
| clippy | clean |

> Note: the generated `main.rs` differs run-to-run regardless of this change (planner join
> ordering is HashMap-iteration dependent), so the regression signal is the **arrangement
> count + output parity**, both of which are stable and unchanged.

## Why it's worth doing

Separate phase-breakdown profiling (FlowLog vs Souffle on polonius/ci/doop) showed that
graph-construction (spinup), peak RSS, and arrangement teardown all scale with the number
of arrangements the planner emits — so `from_info` is the hot path for keeping that count
down. Keeping it small and free of sync hazards is worth the tidy-up.

### DIFF
--- a/crates/flowlog-build/src/planner/transformation.rs
+++ b/crates/flowlog-build/src/planner/transformation.rs
@@ -218,6 +218,13 @@ fn resolve_fp(fp_map: &HashMap<u64, u64>, fp: u64) -> u64 {
     fp_map.get(&fp).copied().unwrap_or(fp)
 }
 
+/// Build a collection from a fingerprint, hierarchical name, and key/value
+/// layout. Centralizes the `Arc<Collection>` boilerplate that every
+/// transformation constructor would otherwise repeat per input and output.
+fn collection_of(fp: u64, name: &str, layout: &KeyValueLayout) -> Arc<Collection> {
+    Arc::new(Collection::new(fp, name.to_string(), layout.key(), layout.value()))
+}
+
 // ========================
 // Constructors
 // ========================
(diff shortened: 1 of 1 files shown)


hdz284:
Closing in favour of #155, which consolidates the three overlapping `from_info` cleanups (#153, #154, #155).

#155 keeps this PR's `collection_of` helper verbatim. The one difference worth calling out: this PR moves the `tag` computation *into* each constructor but keeps it as a **separate `match` from the variant dispatch** right below it — so the "tag mirrors the variant" coupling is relocated rather than removed (the two switches over `(is_row_input, is_row_output)` can still drift). #155 instead yields `(tag, ctor)` from a single match arm, so the tag and the variant it seeds come from the same place and can't drift. Same byte-identical fingerprints / parity story as here.

PR #155, opened 2026-06-12
refactor(planner): one source of truth for the fingerprint tag↔variant pairing + factor Collection building (consolidates #153, #154)

hdz284:
## Summary

A readability/robustness pass over the content-canonical fingerprint +
arrangement-sharing path introduced by #148 (`Transformation::from_info` →
within-stratum dedup → recursion factoring → cross-stratum prune → codegen
`arranged_map` reuse).

> **Consolidates #153 and #154** (three of us opened overlapping cleanups of the
> same `from_info` within minutes). This PR is the union of the parts that hold
> up; the other two are closed in its favour — see *Relationship to #153/#154*
> below.

## What changed (one file, net −25 lines)

1. **One source of truth for the tag↔variant pairing.** `from_info` decided the
   fingerprint `tag` string in one `match`, while the enum variant was decided
   by a *second* `match` inside `kv_to_kv`/`join`/`antijoin`. The soundness
   property *"equal fingerprint ⇒ same variant"* was held only by convention
   across those two sites — its own comment flagged it (`// Tag mirrors the
   variant picked below`). Each constructor now yields `(tag, ctor)` from a
   **single** match arm, so the tag and the variant it stands for cannot drift;
   `from_info` drops its redundant tag match and just dispatches.
2. **Factor the `Collection` boilerplate.** The 6 hand-rolled
   `Arc::new(Collection::new(fp, name.to_string(), layout.key(), layout.value()))`
   copies collapse into one `collection_of(fp, name, layout)` helper.

### Provably no behaviour change
The tag strings and the `compute_fp` inputs — `(variant tag, resolved input
fps, flow)` — are byte-for-byte unchanged, so every content fingerprint (and
therefore all sharing / dedup / prune decisions and the generated dataflow) is
identical:

- **Generated code content-identical** (sorted, to factor out codegen's
  pre-existing statement-order nondeterminism) for the deterministic programs
  `doop.dl` and `polonius_str.dl` (0 diff).
- **Arrangement counts unchanged**: doop `192`, polonius `56`,
  context-insensitive `~1177–1178` (this program's count already varies ±1
  run-to-run from its `ord()` relations, independent of this change).
- `218` `flowlog-build` unit tests + `111/111` e2e lib fixtures pass; `clippy` clean.

## Relationship to #153 / #154

All three did the same core fix (co-locate the tag with the variant). The
differences, and why this is the union worth keeping:

| | tag↔variant | `Collection` factoring | module docs |
|---|---|---|---|
| #153 | co-located | `collection()` + `binary_collections()` | **added to `planner/mod.rs`** — @HarukiMoriarty flagged: *"optimization tricks should not be put at the crate entry level"* |
| #154 | tag moved into the constructor but still a **separate** match from the variant dispatch — the drift hazard persists, just relocated | `collection_of()` | none |
| **this** | **single `(tag, ctor)` match** — the hazard is structurally gone | `collection_of()` (same as #154) | none (honours the #153 review feedback) |

So this PR takes the only tag↔variant form that actually removes the hazard
(`(tag, ctor)`), keeps #154's `collection_of`, and deliberately omits the
module-entry docs #153 was asked to reconsider.

## Side investigation (no code here): order-independent filter hashing

While in here I also tested the obvious follow-on — making the content
fingerprint hash conjunctive filters order-independently (the closed #145 idea:
hash `flow.content_key()` with key/value positional but constraints/compares/UDF
predicates as sorted multisets) — and **benchmarked it to a negative result**:

- arrangements: context-insensitive (flattened DOOP) `1178 → 1167` (−11);
  **neutral** on every hand-written example (doop, polonius, cspa, csda, galen…).
- results byte-identical on doop@batik (26/26), context-insensitive@batik
  (21/21 row counts), polonius@clap (20/20).
- **but solve time `53.5s → 54.5s` (≈ +1.9%, w16, 5 clean interleaved runs)** on
  context-insensitive, RSS flat — every head sample exceeded every base sample.

Cause: order-independent dedup keeps *one* rule's filter order as the shared
representative and discards the planner's per-rule selectivity ordering (SIP /
pushdown); a less-selective predicate first in the hot `&&` over ~28M tuples
outweighs the saved arrangements. **Conclusion: not worth it — #148's
order-sensitive flow hash is the right call.** Left out of this PR; recording it
so it isn't re-attempted.

## Validation
- `cargo test --release --workspace`: **277 passed, 0 failed**
- e2e lib fixtures: **111/111**
- `clippy`: clean
- real-data parity above on batik (DOOP DaCapo) and clap (polonius)

Base: `main-next`.

### DIFF
--- a/crates/flowlog-build/src/planner/transformation.rs
+++ b/crates/flowlog-build/src/planner/transformation.rs
@@ -218,6 +218,28 @@ fn resolve_fp(fp_map: &HashMap<u64, u64>, fp: u64) -> u64 {
     fp_map.get(&fp).copied().unwrap_or(fp)
 }
 
+/// Build a collection handle from a fingerprint, hierarchical name, and
+/// key/value layout. Centralizes the `Arc<Collection>` boilerplate that every
+/// transformation constructor would otherwise repeat for each input and output.
+fn collection_of(fp: u64, name: &str, layout: &KeyValueLayout) -> Arc<Collection> {
+    Arc::new(Collection::new(
+        fp,
+        name.to_string(),
+        layout.key(),
+        layout.value(),
+    ))
+}
+
+/// Builds a unary [`Transformation`] variant from its input, output, and flow.
+/// Chosen *together* with the variant's fingerprint tag in a single match arm
+/// (see [`Transformation::kv_to_kv`]) so the two can never drift apart.
+type UnaryCtor = fn(Arc<Collection>, Arc<Collection>, TransformationFlow) -> Transformation;
+
+/// Builds a binary (join / antijoin) [`Transformation`] variant. As with
+/// [`UnaryCtor`], picked alongside its fingerprint tag in one match arm.
+type BinaryCtor =
+    fn((Arc<Collection>, Arc<Collection>), Arc<Collection>, TransformationFlow) -> Transformation;
+
 // ========================
 // Constructors
 // ========================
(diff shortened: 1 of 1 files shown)


hdz284:
Superseded by #157, which consolidates all the sharing/factoring findings (this PR's content included) into one clean, fully-tested PR.

PR #157, opened 2026-06-12
planner: tidy the arrangement-sharing path + document & test cross-stratum factoring

hdz284:
**Supersedes #152, #153, #154, #155** — one PR for all the sharing/factoring
findings. Docs + tests + a behaviour-preserving refactor; **no engine change**.

## 1 · Refactor `Transformation::from_info`  (was #153 / #154 / #155)

- Pick the fingerprint `tag` and the enum variant in **one** match arm
  (`(tag, ctor)`), so *"equal fingerprint ⇒ same variant"* is structural rather
  than convention — it used to be two parallel matches kept in sync by a comment.
- Factor `collection(..)` + `binary_collections(..)` out of the six hand-rolled
  `Collection::new` copies.

`compute_fp` inputs (tag strings, resolved input fps, flow) are byte-identical, so
all content fingerprints, sharing, and generated code are unchanged.

## 2 · Document the sharing pipeline  (new)

`planner` module overview: materialize → within-stratum dedup → recursion
factoring → cross-stratum prune. Plus a recorded **negative result** —
order-independent filter hashing costs ~2% solve on the only programs it helps
(#155), so it's deliberately not done.

## 3 · Cross-stratum factoring: soundness + coverage  (was #152)

Rule-level stratification splits a relation's *producer* rules across strata, so
it is **not** "fixed" after a stratum; factoring is sound anyway because a
*consumer* is stratified after **all** producers and so reads the completed
relation. The intervening-IDB-update gate is a no-op in `datalog-batch` and
load-bearing only in extended mode — and had **zero test coverage**. Adds the
gate regression + a stratifier unit test + 4 Soufflé-verified fixtures, and
corrects the imprecise "fixed / fully computed" docs.

## Tested

- `cargo test -p flowlog-build` **275 passed**; full lib suite **116 passed**;
  the 5 new fixtures green in **compiler and lib** modes; clippy/fmt clean; no new
  doc warnings.
- **No behaviour change**: arrangement counts unchanged (doop 192, polonius 56,
  context-insensitive batik 190); doop @ batik (`-w16`) 107.5 s / 52.9 GB = baseline.
- **Gate is load-bearing**: deleting it leaves all existing tests green but fails
  the new `arrangement_share_fixpoint_idb` fixture (R2 2/3, R3 3/6 rows).

Leaves #138 (busy-spin parking) alone — an unrelated perf change.


hdz284:
Reviewed this consolidation against the prior cluster (#152–#155) and `main-next`. Good to have the four in one place — the `from_info` refactor (§1) is the right `(tag, ctor)` + `collection(..)` form (the tag↔variant drift hazard is structurally gone), and §3 adds the load-bearing gate tripwire that was missing. Two things to fix before merge:

**1. §2 re-introduces the exact thing @HarukiMoriarty asked to reconsider on #153.**
The new `planner/mod.rs` header carries the full sharing-pipeline walkthrough *and* a `## Rejected: order-independent filter hashing` optimization-trick section — at the crate-module entry. That is precisely the #153 review feedback:

> the comments cleaning should be in a more careful way, especially **optimization tricks should not be put at the crate entry level**, that will significantly increase the complexity of understanding.

Suggest: keep `planner/mod.rs` to a 1–2 line "what the planner does" blurb; the per-pass detail is already documented locally (`Transformation::from_info`, `StratumPlanner`, `prune_cross_stratum_duplicates`), and the "rejected order-independent hashing" note belongs in a short code comment next to the flow hashing (or just this PR description) — not at the crate entry.

**2. Reconcile the "context-insensitive" / "doop" metrics — they mix programs under one label.**
- `doop 192` ✓ matches `example/program_analysis/doop.dl`.
- §2's negative result cites `−11 arrangements` on the **generic-component-flattened** context-insensitive program, which emits **~1177–1178** `arrange_by_*` (1178→1167) — not the `context-insensitive batik 190` in the test table. Those are two different programs both labelled "context-insensitive"; please name them distinctly so the −11 / ~2% isn't read as applying to the 190-arrangement run.
- The perf line `doop @ batik -w16 107.5 s / 52.9 GB` is ~8× / ~12× the `example/doop.dl` profile measured here (≈13 s / 4.4 GB at -w16, VarPointsTo 37,237,353). If that row is a different/full program, label it; otherwise it looks off.

Net: the consolidation itself is great — just (1) honour the crate-entry-docs feedback and (2) make the program labels unambiguous. (FWIW the stale branches behind the now-closed #152/#153/#154/#155 can be deleted too.)

hdz284:
Pushed b46fa57 applying **item 1**: trimmed `planner/mod.rs` back to a short overview that points at the per-pass docs (`Transformation::from_info`, `StratumPlanner`, `ProgramPlanner`), and dropped the `## Rejected: order-independent filter hashing` optimization-trick section from the crate entry per @HarukiMoriarty's #153 feedback. The rejected-optimization finding stays in this PR's description. Docs-only; 219 `flowlog-build` tests pass, clippy clean, no doc warnings introduced.

**Item 2** (metric labels) is left for you to confirm since I can't tell which programs the `context-insensitive batik 190` arrangement row and the `doop @ batik 107.5 s / 52.9 GB` perf row came from — they don't match what I measured here for the repo programs (`example/doop.dl` ≈ 192 arrangements / ≈13 s / 4.4 GB at -w16; the generic-component-flattened context-insensitive ≈ 1177–1178 arrangements, which is the 1178→1167 program §2 cites). Worth labelling those two rows distinctly.

HarukiMoriarty:
Leave for further structured data refactoring.

PR #177, opened 2026-06-13
fix(compiler): warn when vm.max_map_count is too low for mimalloc on large runs

hdz284:
## Summary

Large analyses can crash with a cryptic `memory allocation of <N> bytes failed` (SIGABRT) **even when the machine has hundreds of GB of RAM free**, producing zero output and no `Dataflow executed` stamp. This is not an OOM and not a correctness bug — it's the generated binary's **mimalloc** allocator exhausting the default Linux mapping ceiling **`vm.max_map_count=65530`**.

Found while benchmarking context-insensitive points-to (`-w 32`) on the DaCapo **jython** program: FlowLog aborts at ~137 GB resident; the same run completes correctly (`VarPointsTo = 371,435,757`) once `vm.max_map_count` is raised.

## Root cause (measured)

mimalloc keeps a low mapping count for most of a run, then its purge/decommit splits arenas into tens of thousands of VMAs as the working set peaks. Sampling `/proc/<pid>/maps` live:

| RSS | VMAs |
|----:|-----:|
| 109 GB | 445 |
| 120 GB | 44,982 |
| 122 GB | **65,530** (= `vm.max_map_count`) |
| 130 GB | 65,531 → **mmap ENOMEM → SIGABRT** |

- Reproduced 3× at the default limit; **completes** at `vm.max_map_count=1048576`.
- mimalloc option tuning does **not** fix it: `MIMALLOC_PURGE_DELAY=-1` and `MIMALLOC_PURGE_DECOMMITS=0` avoid the *solve*-phase split but just move the VMA explosion into the *output* phase (and raise peak RSS to ~156 GB). Raising the kernel limit is the only robust fix.

The mappings ceiling is a system-wide sysctl the binary cannot raise itself, so this PR makes the failure **discoverable and preventable** rather than silently fatal.

## Changes

- **`flowlog-compiler` codegen** — generated `main()` (batch + incremental) prints a one-line, Linux-only startup warning when `vm.max_map_count` is still at the kernel default (`<= 65536`), pointing at the fix. Zero allocator-behavior change; fires only on untuned kernels (à la Redis/Elasticsearch startup checks).
- **`env/env.sh`** — one-time setup now raises and persists `vm.max_map_count=1048576` (`/etc/sysctl.d/99-flowlog.conf`) on Linux.
- **`README.md`** — documents the requirement and the symptom.

## Validation

- `cargo test --workspace` green (218 + suites, 0 failed); `cargo fmt --check`, `cargo clippy`, `typos` clean.
- Generated binary at default limit (`65530`): prints the warning **and still runs**; at `1048576`: silent. Souffle (glibc/OpenMP) on the same workload stays at ~270 VMAs and is unaffected — this is mimalloc-specific.

> Note: only a *warning* — a process can't raise a system-wide sysctl. It converts an undebuggable SIGABRT into an actionable message; `env/env.sh` + the README provide the actual fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,13 +62,18 @@ Flag reference: [Compiler CLI](#compiler-cli). For incremental mode and the prof
 
 ## System requirements
 
-FlowLog runs on a stock Linux, macOS, or Windows host, but large analyses — many
-worker threads over hundreds of millions of tuples — can run into OS-level
-limits. On Linux especially, the generated binary's allocator (mimalloc) maps a
-great many memory regions, so a low `vm.max_map_count` can make a big run abort
-with `memory allocation of <N> bytes failed` even when plenty of RAM is free.
-Before a large run, review the host settings (including `vm.max_map_count`) in
-the setup guide: <https://www.flowlog-rs.com/tutorial/getting-started/system-config>.
+FlowLog's performance and stability depend on a number of host and OS-level
+settings. For example, on Linux a large analysis can abort with `memory
+allocation of <N> bytes failed` even when memory is plentiful, because the
+allocator maps many memory regions and exhausts a low
+`vm.max_map_count`. Raising it resolves this:
+
+```console
+$ sudo sysctl -w vm.max_map_count=1048576
+```
+
+Before running, review the recommended host configuration in the setup guide:
+<https://www.flowlog-rs.com/tutorial/getting-started/system-config>.
 
 ## Architecture
 
--- a/README.md
+++ b/README.md
@@ -60,6 +60,16 @@ $ ./reach_bin -w 4
 
 Flag reference: [Compiler CLI](#compiler-cli). For incremental mode and the profiler, see <https://www.flowlog-rs.com/>.
 
+## System requirements
+
+FlowLog runs on a stock Linux, macOS, or Windows host, but large analyses — many
+worker threads over hundreds of millions of tuples — can run into OS-level
+limits. On Linux especially, the generated binary's allocator (mimalloc) maps a
+great many memory regions, so a low `vm.max_map_count` can make a big run abort
+with `memory allocation of <N> bytes failed` even when plenty of RAM is free.
+Before a large run, review the host settings (including `vm.max_map_count`) in
+the setup guide: <https://www.flowlog-rs.com/tutorial/getting-started/system-config>.
+
 ## Architecture
 
 A `.dl` program compiles through five stages; three side modules assist the planner and codegen:
--- a/flowlog-compiler/src/assembly/batch.rs
+++ b/flowlog-compiler/src/assembly/batch.rs
@@ -43,8 +43,11 @@ pub(crate) fn gen_batch_main(
         ..
     } = input;
 
+    let max_map_guard = super::gen_max_map_count_guard();
+
     quote! {
         fn main() {
+            #max_map_guard
             let args: Vec<String> = std::env::args().collect();
             let barrier = worker_barrier_from_args(&args);
 
--- a/flowlog-compiler/src/assembly/batch.rs
+++ b/flowlog-compiler/src/assembly/batch.rs
@@ -43,11 +43,8 @@ pub(crate) fn gen_batch_main(
         ..
     } = input;
 
-    let max_map_guard = super::gen_max_map_count_guard();
-
     quote! {
         fn main() {
-            #max_map_guard
             let args: Vec<String> = std::env::args().collect();
             let barrier = worker_barrier_from_args(&args);
 
--- a/flowlog-compiler/src/assembly/inc.rs
+++ b/flowlog-compiler/src/assembly/inc.rs
@@ -41,8 +41,11 @@ pub(crate) fn gen_incremental_main(
         ..
     } = rp;
 
+    let max_map_guard = super::gen_max_map_count_guard();
+
     quote! {
         fn main() {
+            #max_map_guard
             let args: Vec<String> = std::env::args().collect();
 
             let shared_txn: Arc<RwLock<TxnState>> =
--- a/flowlog-compiler/src/assembly/inc.rs
+++ b/flowlog-compiler/src/assembly/inc.rs
@@ -41,11 +41,8 @@ pub(crate) fn gen_incremental_main(
         ..
     } = rp;
 
-    let max_map_guard = super::gen_max_map_count_guard();
-
     quote! {
         fn main() {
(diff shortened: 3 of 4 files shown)


HarukiMoriarty:
This is a system setting and should be optional for users, I already flagged it in the documentations and should not be inside of flowlog: https://www.flowlog-rs.com/tutorial/getting-started/system-config.

hdz284:
Good call — thanks, and agreed that flowlog shouldn't be changing system settings. I've revised:

- **Dropped** the `env/env.sh` auto-tuning (it forced a system setting — exactly the "should be optional" concern) and the README duplication (your [system-config doc](https://www.flowlog-rs.com/tutorial/getting-started/system-config) already covers it).
- **Kept** only a one-line, Linux-only startup **diagnostic** in generated `main()`. It changes nothing — it just fires when `vm.max_map_count` is still at the kernel default and points at your doc, so the otherwise cryptic `SIGABRT` / `memory allocation of <N> bytes failed` (zero output, looks like a spurious OOM) becomes actionable.

Diff is now 3 files / +39, codegen-only. If you'd rather not have any diagnostic in the binary either, I'm happy to drop it and close this in favor of the docs — your call.

For context, this surfaced on context-insensitive points-to × **jython** at `-w 32`: the abort hit at ~137 GB with 486 GB free; the per-process VMA count climbed 445 → 65,530 (the default cap) then `mmap` failed. With the cap raised it completes (VarPointsTo = 371,435,757), ~3.9× faster than Soufflé and at lower peak RSS.

HarukiMoriarty:
The investigation here is excellent — the VMA-vs-RSS table, the 3× repro, and ruling out `MIMALLOC_PURGE_*` tuning make the root cause airtight. **`env/env.sh` raising `vm.max_map_count` and the README are the real fix and I'm happy to land those.**

My reservation is narrower: the **unconditional warning embedded in every generated `main()`**. A few concrete reasons, not just taste:

1. **It keys off a static precondition, not the actual failure.** The guard fires whenever `limit <= 65536`, which is the default on essentially every stock Linux box — so every generated binary prints this on every startup, including the overwhelming majority of runs that are small and will never approach 65530 VMAs. The abort needed a ~137 GB / `-w 32` jython run. A warning that fires on ~all runs is one operators quickly learn to ignore — exactly when a genuinely large run needs them to read it.

2. **Altitude / coupling.** The message is only correct because the generated binary uses mimalloc (wired up in `flowlog-compiler/src/imports.rs`). Inlining `/proc`-reading + a hard-coded 65536 + a mimalloc-specific message into emitted `main()` couples the language artifact to one allocator and one kernel tunable, and bakes it in — it can't be silenced without recompiling.

3. **The Redis/Elasticsearch precedent doesn't quite transfer.** Those are long-lived daemons where the check runs once per process lifetime and the tunable matters for any production deploy. A flowlog binary is closer to a CLI invoked repeatedly (tests, CI, incremental txns), so unconditional stderr on every start is noise.

4. **On the supported path it never fires** — `env.sh` raises the limit, so the warning only triggers for users who skipped setup, which is what the README already covers.

### Suggested alternative (if we want to keep a proactive signal)

Gate it on a large-run hint and factor the logic out of the two `main()` bodies into one helper, so the threshold/message live in one testable place and small/CI runs stay quiet:

```rust
/// Warn once at startup only when a large, highly-parallel run could plausibly
/// exhaust the Linux mapping ceiling under mimalloc. Small runs stay silent.
#[cfg(target_os = "linux")]
pub fn warn_if_low_max_map_count(workers: usize) {
    const WORKER_HINT: usize = 16;   // small runs never approach the VMA ceiling
    const LOW_LIMIT: u64 = 65_536;
    if workers < WORKER_HINT { return; }
    let Ok(raw) = std::fs::read_to_string("/proc/sys/vm/max_map_count") else { return };
    let Ok(limit) = raw.trim().parse::<u64>() else { return };
    if limit <= LOW_LIMIT {
        eprintln!(
            "flowlog: warning: vm.max_map_count = {limit} is at the kernel default; \
             large analyses can exhaust per-process memory mappings and abort with \
             \"memory allocation of <N> bytes failed\" even with ample free RAM. \
             See https://www.flowlog-rs.com/tutorial/getting-started/system-config"
        );
    }
}
#[cfg(not(target_os = "linux"))]
pub fn warn_if_low_max_map_count(_workers: usize) {}
```

Generated `main()` then becomes a single gated call passing the parsed worker count, instead of an inlined block in both batch and incremental.

Two reasonable homes:
- **`flowlog-runtime`** — cleanest separation and independently testable, but note it ships in the manually-pinned runtime contract, so it'd need a runtime release + scaffold pin bump.
- **a small codegen helper** (`assembly/mod.rs`) emitted once and called once — keeps it on the compiler's release cadence, no pin dance.

Honestly I'd also be fine **dropping the in-binary warning entirely** and relying on `env.sh` + README, since the supported path already avoids the failure. Either way, I don't think we should merge the unconditional codegen-embedded `eprintln` as-is. What do you think?

HarukiMoriarty:
Two things on closer look:

1. **This PR's diff is only the codegen warning** (`+39/-0` across `assembly/{batch,inc,mod}.rs`). The `env/env.sh` and `README` items in the description aren't actually in it — `env.sh` doesn't touch `vm.max_map_count` and the README has no system-config note on `main-next`. So dropping the warning empties the PR.

2. Since the [system-config docs page](https://www.flowlog-rs.com/tutorial/getting-started/system-config) already exists, let's not put a check in the binary at all. Instead just add a short **"System requirements"** pointer in the README that tells users to be aware of host settings (incl. `vm.max_map_count`) and links to that page.

hdz284:
Done in 592562e — took your recommendation to drop the in-binary check entirely.

- **Removed** `gen_max_map_count_guard` and its two `main()` call sites (the unconditional `eprintln!` fired whenever `vm.max_map_count <= 65536`, i.e. on ~every stock box, and coupled emitted `main()` to mimalloc + one kernel tunable).
- **Added** a short **System requirements** section to the README pointing at the [system-config guide](https://www.flowlog-rs.com/tutorial/getting-started/system-config), noting host settings incl. `vm.max_map_count`.

Net for the mimalloc change: −39 codegen / +10 README; `cargo build --release` clean.

Heads-up unrelated to this change: the branch currently also carries three renovate dep-bump commits (#178/#179/#180) that aren't on `main-next`. Happy to rebase those out if you'd prefer this PR be README-only.

HarukiMoriarty (approved):
LGTM.

PR #200, opened 2026-07-02
Perf(codegen): elide identity projections (drop no-op flat_maps)

hdz284:
## What & why

A `RowToRow` / `RowToKv` transformation whose output reproduces the input row **verbatim** — every column, exactly once, in order, with no predicate — lowers to `flat_map(|row| std::iter::once(row))`. That is a **no-op differential-dataflow operator**: the input collection already *is* the output.

This PR detects that case and aliases the input collection instead (`let out = inp.clone();`), so the operator right after (typically the `arrange_by_self` registered in the same arm, or the next consumer) reads the input directly.

Motivation: every `worker.step()` schedules the **whole** dataflow regardless of how little data actually moved, so each surviving operator costs fixed per-round scheduling. On request/response-shaped workloads (many tiny updates over a resident program — e.g. per-query authorization) that fixed cost dominates, so dropping provably-dead operators is a direct, program-agnostic win. It also trims dataflow assembly for large programs.

## The change

`flowlog-build/src/codegen/flow/transformation.rs`:
- `is_identity_row_projection(args, row_arity)` + `is_plain_row_column_ref(arg, idx)` — detect a full, in-order, plain-variable projection of every input row column (no arithmetic, cast, constant, reorder, or dropped/added column).
- `RowToKv` arm: when `pred.is_none() && output.is_k_only() && flow.value().is_empty() && is_identity_row_projection(flow.key(), input.arity().1)`, emit `let out = inp.clone();` instead of the `flat_map`; the arrangement registration is unchanged.
- `RowToRow` arm: analogous, guarded on `flow.key().is_empty()`.

**Behavior-preserving by construction:** the guards mean the removed operator is provably the identity (`|row| once(row)`), so aliasing is exact. Any predicate, cast, arithmetic, reorder, or projection that drops/duplicates a column falls through to the original `flat_map`.

## Effect (elided operators)

Generated-code `flat_map` counts, `--mode datalog-batch`:

| program  | before | after | Δ |
| -------- | -----: | ----: | ---: |
| doop     | 214    | 194   | −20 |
| reach    | 4      | 2     | −2  |
| pointsto | 9      | 7     | −2  |
| tc       | 4      | 3     | −1  |
| galen    | 13     | 12    | −1  |
| andersen | 6      | 5     | −1  |
| cspa     | 12     | 11    | −1  |
| csda     | 4      | 3     | −1  |

(and 5 on a sasy request/response policy skeleton, 12 → 7.)

## Validation

- `cargo test -p flowlog-build` — **80 pass** (73 existing + **7 new** unit tests for the detector: in-order identity, key/value-flag-ignored-for-rows, reorder, drop/duplicate, arity mismatch, empty/zero-arity, arithmetic-tail-not-plain).
- Oracle suite (`make oracle`, batch mode): **andersen** (2.65M tuples) and **cspa** (88.9M tuples) byte-identical to the Soufflé reference with the patched compiler.
- sasy incremental integration: `|deny|` decision identical at every turn (cross-checked against Soufflé + a hand-written DD oracle); per-query latency unchanged/within noise.
- `cargo +nightly fmt --check` and `cargo clippy` clean.

## Scope / follow-ups

- Limited to the two unary arms that emit an identity `flat_map` before an arrangement (`RowToKv`, `RowToRow`) — the cases observed in generated code. The same idea extends to `KvToKv` self-projections; left out here to keep the change small and obviously correct.
- Complementary, not yet included: collapsing an idempotent `threshold` stacked on an already-canonical collection (e.g. the double-distinct on antijoin/`authorized`).

### DIFF
--- a/flowlog-build/src/codegen/flow/transformation.rs
+++ b/flowlog-build/src/codegen/flow/transformation.rs
@@ -23,8 +23,11 @@ use crate::codegen::arg::compute_kv_param_tokens;
 use crate::codegen::arg::row_pattern_and_fields;
 use crate::codegen::data_type_tokens;
 use crate::codegen::ident::find_local_ident;
+use crate::planner::ArithmeticArgument;
+use crate::planner::FactorArgument;
 use crate::planner::StratumPlanner;
 use crate::planner::Transformation;
+use crate::planner::TransformationArgument;
 
 impl CodeGen {
     /// Generate differential dataflow pipelines for a single transformation.
@@ -75,16 +78,6 @@ impl CodeGen {
                 let inp = find_local_ident(local_fp_to_ident, input.fingerprint());
                 let out = find_local_ident(local_fp_to_ident, output.fingerprint());
 
-                // Profiler hook (optional)
-                with_profiler(profiler, |profiler| {
-                    profiler.map_join_operator(
-                        transformation_name,
-                        vec![inp.to_string()],
-                        out.to_string(),
-                        output.fingerprint(),
-                    );
-                });
-
                 // Type inference + row pattern
                 let input_arity = input.arity().1;
                 let (row_pat, row_fields) = row_pattern_and_fields(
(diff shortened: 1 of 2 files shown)


HarukiMoriarty (approved):
LGTM.

hdz284:
Force-pushed to update this PR (was based on the pre-#205 `main-next`). Two things changed:

1. **Rebased onto current `main-next`** (now includes #205 n-ary concatenate fusion + #201 threshold_total), as required by branch protection (`strict_up_to_date`). Clean replay, no conflicts — the identity-elision diff is unchanged.

2. **Added one commit fixing a profiler regression this PR introduced.** Eliding the identity `flat_map` for a copy rule (`B :- A`) also dropped its profiler node, because it was gated behind `if !is_identity`. Any program with a copy rule then failed profiler-model resolution under `-P` — `flowlog-visualizer` aborted with *"references fingerprint '…' not found in any node"* as soon as a downstream rule (`C :- B`) referenced the elided relation's fingerprint.

   Fix: register a dedicated **0-operator** alias node (`Profiler::identity_alias_operator`) for the elided relation. It carries the output fingerprint so downstream lookups resolve, but advances the operator-address counter by 0 (no timely operator exists to attribute), so the address model stays exact. Dataflow output is unchanged — profiler-only.

Validation: full workspace builds clean; 257 lib tests pass (incl. this PR's 5 `identity_projection_tests`); `-P` + `flowlog-visualizer` now exit 0 with a report produced on copy-rule programs — plain copy, chained copies, copy→join, single-column copy, and recursive TC with copy rules — all with unchanged results. Flagging for a re-review since code changed after approval.

PR #201, opened 2026-07-02
Perf(codegen): use `threshold_total` for non-recursive dedup in incremental mode

hdz284:
## What & why

Incremental modes carry **`i32` diffs**, so enforcing set semantics means clamping multiplicities to 0/1. (`consolidate` alone is only set-correct under the **idempotent `Present` semiring** that `datalog-batch` uses — `P + P = P`; with `i32`, `P + P = 2P`, and the extra multiplicity leaks into downstream joins / negation / aggregation.) So every incremental dedup site emitted the general `Collection::threshold`, which lowers to `arrange_by_self` + **`reduce_abelian`** — the partial-order reduce machinery, maintained across every commit.

But the **outer timestamp in incremental mode is `u32`, a total order**. For **non-recursive** and **antijoin** dedup — which run in that outer scope — `ThresholdTotal::threshold_total` computes exactly the same result via a lighter, total-order-specialised path. **Recursive** dedup runs under `Product<_, _>` (only a partial order, not `TotalOrder`), where `threshold_total` doesn't apply, so it keeps the general `threshold`.

## The change

`flowlog-build/src/codegen/dedup.rs`:
- `dedup_nonrecursive` (inc) and `dedup_antijoin` (inc): emit `.threshold_total(|_, w| if *w > 0 { SEMIRING_ONE } else { 0 })` and mark the `threshold_total` feature (so `ThresholdTotal` is imported).
- `dedup_recursive` (inc): unchanged — general `threshold` (Product time is only a partial order).
- Batch codegen: untouched.

## Efficiency

sasy request/response skeleton, single worker, engine-internal per-commit time (median of many interleaved runs):

| turns | general `threshold` | `threshold_total` | speedup |
| ----: | ------------------: | ----------------: | ------: |
|    96 |            361.7 ms |          279.7 ms | **+22.7%** |
|   192 |           1116.6 ms |          926.3 ms | **+17.0%** |
|   384 |           4670.8 ms |         4239.2 ms | **+9.2%**  |

**40/40** runs faster at T=96 (±0.3 ms SE). Memory roughly neutral (RSS within run-to-run noise). The win is largest at short conversations — the operating point of this workload — because at large T the maintained-slice recursion (still on the general `threshold`) increasingly dominates.

## Correctness

- sasy incremental integration: `|deny|` identical at every turn vs Soufflé **and** a hand-written DD oracle, across **40 random graphs** (arbitrary edges, cycles, multi-current).
- `reach` (recursive transitive closure) in incremental mode: **byte-identical** output vs the general `threshold` build — exercises the recursive (kept-`threshold`) + non-recursive (`threshold_total`) split.
- Full workspace suite green (**302 tests**); updated the dedup unit test to assert the new inc mapping (non-recursive/antijoin → `threshold_total`, recursive → `threshold`).

## Notes on the operator choice (for reviewers)

Summary of "which dedup where" in incremental mode:
- **`threshold_total` / `distinct_total`** — total-order scopes (non-recursive, antijoin). Cheapest correct option. *(this PR)*
- **`threshold` / `distinct`** — required inside recursion (`Product` partial order); `distinct` == `threshold` cost-wise (both general `reduce_abelian`).
- **`consolidate`** — set-incorrect in inc (i32 not idempotent); correct only where multiplicity is provably ≤ 1, or in batch (`Present`).
- **`threshold_semigroup`** — the total-order primitive `threshold_total` is built on; batch-recursive uses it directly.


HarukiMoriarty (changes requested):
The `threshold_total` direction looks good — the outer inc timestamp is a total order and the numbers back it up. `dedup_nonrecursive` is fine as-is.

One concern with the `dedup_antijoin` change though: antijoins don't always run in the outer scope. Stratified negation is legal *inside* a recursive rule (negating a lower-stratum relation), and in that case the antijoin — including its dedup — is emitted inside the `iterate` scope, where the timestamp is `Product<u32, u16>`, which is only a partial order. `threshold_total` requires `TotalOrder`, so the generated crate fails to compile.

Minimal repro (`--mode datalog-inc --check`):

```datalog
.decl seed(x: number)
.decl edge(x: number, y: number)
.decl blocked(x: number)
.decl reach(x: number)
.input seed(IO="file", filename="seed.csv", delimiter=",")
.input edge(IO="file", filename="edge.csv", delimiter=",")
.input blocked(IO="file", filename="blocked.csv", delimiter=",")
.output reach
reach(x) :- seed(x).
reach(y) :- reach(x), edge(x, y), !blocked(y).
```

On this branch:

```
= note: the following trait bounds were not satisfied:
        `timely::order::Product<u32, u16>: TotalOrder`
        which is required by `Collection<'_, timely::order::Product<u32, u16>, ...>: ThresholdTotal<...>`
error: could not compile `recneg` (bin "recneg") due to 2 previous errors
```

Same program compiles clean on `main-next`, so this is a regression from the `dedup_antijoin` half of the change.

Suggested fix: make the antijoin dedup scope-aware, mirroring the existing recursive/non-recursive split — outer-scope antijoins keep `threshold_total` (the perf win stands), antijoins inside a recursive stratum fall back to the general `threshold`.

Separately, this slipped through because we have no incremental fixture combining recursion + negation — none of the `datalog-inc` fixtures exercises that shape. Worth adding one (the repro above works) under `tests/fixtures/datalog-inc/` so this path is covered going forward.

HarukiMoriarty:
also, please sign on the commit to pass DCO checks.

hdz284:
Thanks @HarukiMoriarty — fixed in 6168bf0, and signed off both commits (DCO green now).

**Scope-aware antijoin dedup.** You were exactly right: the antijoin dedup can't unconditionally use `threshold_total`, because stratified negation is legal *inside* a recursive stratum, where the antijoin runs under `Product<u32, u16>` (only a partial order). `gen_transformation` now threads a `recursive` flag from its two call sites (`recursive.rs` → `true`, `non_recursive.rs` → `false`), and `dedup_antijoin(recursive)` keeps `threshold_total` in the outer scope (the perf win stands) but falls back to the general `threshold` inside recursion — mirroring the existing `dedup_recursive`/`dedup_nonrecursive` split. `dedup_nonrecursive` is unchanged.

**Fixture.** Added `tests/fixtures/datalog-inc/recursive_neg_delta/` — recursion + negation (`Reach(y) :- Reach(x), Edge(x, y), !Blocked(y)`), the shape no incremental fixture covered. I verified it reproduces the `Product<u32, u16>: TotalOrder` error on the parent commit and passes here; blocking/unblocking a node ripples reachability through the recursion across three transactions (`+1` cascade on unblock, `-1` cascade on re-block).

Full suite green locally: 73 `flowlog-build` unit tests + all 122 binary-mode fixtures.

HarukiMoriarty (changes requested):
Thanks for the antijoin fix — `dedup_antijoin(recursive)` threading the flag from the two call sites is exactly right, and `recursive_neg_delta` is a good fixture to add.

Unfortunately the same `Product<_, _>: TotalOrder` regression still fires, through a **second** dedup site this PR didn't make scope-aware. Requesting changes for that.

## The remaining bug

The first commit also switched `dedup_nonrecursive()` to emit `threshold_total` unconditionally — but `dedup_nonrecursive()` is not always non-recursive. It's also called from the `KvToKv` arm of `gen_transformation` (the SIP projection-to-key dedup), which **ignores the new `recursive` flag**. When SIP projects the *recursive* relation to a key, that projection is emitted inside the `iterate` scope (`Product<u32, u16>` time), so `threshold_total` there fails to compile — identical to the antijoin case.

The path is fully reachable:
- SIP builds a `kv_to_kv` with empty value (`is_k_only()`) and `KvPredicates::default()` (no predicate) in `sip.rs` — exactly the `pred.is_none() && output.is_k_only()` condition that emits the dedup in the `KvToKv` arm.
- SIP runs on every rule in a stratum regardless of `is_recursive` (`stratum_planner.rs`), so it fires on recursive rules too.

## Counterexample (`--mode datalog-inc --sip`)

Textbook same-generation:

```datalog
.decl edge(x: number, y: number)
.decl sg(x: number, y: number)
.input edge(IO="file", filename="edge.csv", delimiter=",")
.output sg
sg(x, y) :- edge(x, y).
sg(x, y) :- edge(a, x), sg(a, b), edge(b, y).
```

On this branch (`--check`):

```
error[E0599]: the method `threshold_total` exists for struct
`Collection<'_, Product<u32, u16>, ...>`, but its trait bounds were not satisfied
        `timely::order::Product<u32, u16>: TotalOrder`
```

- Compiles clean on `main-next` — this is a regression from the `dedup_nonrecursive` half of commit 1, not something the fix commit introduced.
- `--sip`-gated: without `--sip` it compiles (no SIP projection is emitted).
- Also reproduces with plain double-recursion (`reach(x,y) :- reach(x,z), reach(z,w), edge(w,y)`) and 3-atom cyclic joins.

## Suggested fix

Same treatment as antijoin: the `KvToKv` dedup site must be scope-aware — outer-scope projections keep `threshold_total`, projections inside a recursive stratum fall back to the general `threshold`. `gen_transformation` already has `recursive` in scope; it just isn't forwarded to this call.

Since this is now the **second** site to need the same flag, consider consolidating rather than threading a bool to a third place later: the operator choice is purely *"is the current scope a total order?"*, so a single scope-aware `dedup` (driven off one flag, or off codegen scope state set/cleared around the `iterate` block) used by all dedup sites would make it impossible to miss one.

## Fixture

`recursive_neg_delta` covers recursion + negation, but no inc fixture exercises recursion under `--sip`. Worth adding a `datalog-inc` fixture that compiles the `sg` program above with `--sip`, so this path is guarded going forward.

hdz284:
Fixed in d9ff58d (signed off, DCO green). You were right on both counts — the second site and the "consolidate it" instinct.

**Root cause was exactly the SIP `KvToKv` projection.** `dedup_nonrecursive()` was doing double duty: the EDB/non-recursive-post-flow dedup *and* the SIP projection-to-key dedup in the `KvToKv` arm. SIP runs on every rule in a stratum, so when it projects the *recursive* relation to a join key the dedup is emitted inside the `iterate` scope (`Product<u32, u16>` time) — `threshold_total` there fails to compile, identical to the antijoin case.

**Consolidated rather than threading a third bool.** Took your suggestion — all three diff-mode dedup sites now route through a single choke point:

```rust
/// Single source of truth for the diff-mode dedup operator: "is the current
/// scope's timestamp a total order?" Outer scope → threshold_total; inside an
/// iterate scope (Product time) → the general threshold.
fn threshold_scoped(&mut self, recursive: bool) -> TokenStream {
    if recursive {
        threshold_nonzero()
    } else {
        self.features.mark_threshold_total();
        threshold_total_nonzero()
    }
}
```

`dedup_setwise` (renamed from `dedup_nonrecursive`, now takes `recursive`), `dedup_recursive`, and `dedup_antijoin` all defer to it, so a new dedup site can't silently pick `threshold_total` and break under `Product` time. The SIP call site forwards the ambient `recursive` flag; the two genuinely-outer sites pass `false`.

**Fixture.** Added `tests/fixtures/datalog-inc/recursive_sip_delta/` — the textbook same-generation program compiled with `--sip`, the `--sip` + recursion shape no inc fixture covered. It reproduces the `Product<u32, u16>: TotalOrder` error on the parent commit; here it compiles and its per-transaction deltas are **byte-identical to the non-SIP oracle** across three transactions (recursive "cousin" derivations `1,4`/`2,3`/`1,5` and a retraction cascade on edge removal). I verified the counterexamples from your review — `sg`, plain double-recursion (`reach(x,y) :- reach(x,z), reach(z,w), edge(w,y)`), and the recursion+neg program — all now compile under `--sip`, and all compile clean without `--sip` and in batch mode.

Full suite green locally: 73 `flowlog-build` unit tests + **all 123** binary-mode fixtures (the two dedup unit tests now also assert the recursive-scope `dedup_setwise` fallback).

HarukiMoriarty (approved):

PR #205, opened 2026-07-04
Perf(codegen): fuse chained unions into a single n-ary concatenate

hdz284:
## What

Wherever the codegen unions `k>=3` collections it emitted a **left-to-right chain of `k-1` binary `concat`s**. This PR replaces every such chain with one **n-ary `Collection::concatenate`** — a single Timely `Concatenate` operator with `k` inputs. Union is associative/commutative, so results (and recursive fixpoints) are identical, but a `k`-way union drops from `k-1` operators to `1`.

Three sites — **every chained union the compiler emits**:

| site | codegen | union |
| --- | --- | --- |
| non-recursive IDB head-union | `non_recursive.rs`, `gen_non_recursive_post_flows` | `head_1 U … U head_k` |
| recursive IDB head-union | `recursive.rs`, `collect_unions` | `head_1 U … U head_k` (feeds the `Variable`) |
| loop `until` gate union | `recursive.rs`, `build_until_conditions` | `base U edge-gated U recursion-gated` |

Binary (`k=2`) unions are left as-is — `concatenate([x]) == concat(x)`, one operator either way — so the only remaining `concat`s are genuine 2-way merges.

## Why

Every `worker.step()` propagates progress through **every** operator, so operator count multiplies Timely's per-step coordination cost, and fewer operators also means a smaller dataflow to construct. It is additionally the **idiomatic DD spelling** of a k-way union, with **no change to the data plane**.

## Correctness (parity-preserving)

- 73 `flowlog-build` unit tests pass; workspace builds clean.
- 3-way non-recursive union and 3-head recursive IDB produce the exact expected relations.
- All four `extend-batch` loop/`until` fixtures (`loop_fixpoint`, `loop_stop_and`, `loop_stop_or`, `loop_stop_relation`) match expected outputs — these exercise both the recursive head-union and the `until`-gate union.
- 9 example programs compile unchanged (`tc`, `sg`, `cc`, `sssp`, `dyck`, `bipartite`, `reach`, `andersen`, `galen`).
- **Profiler (`-P`) kept accurate:** `concat_dedup_operator`'s operator count now reports the single emitted `concatenate` (1, or 0 for a lone head) instead of the pre-fusion `k-1`. That count advances the profiler's per-scope operator-address counter, so leaving it stale would drift every later operator's predicted address; verified on a 3-head union that the fused `concat & dedup` node models 1 `Concatenate` and all following addresses stay aligned.
- Real workload — sasy airline policy (147 rules → ~919 operators, real tau2 trace, 1162 queries): **decision parity Soufflé == FlowLog `1162/1162` on every run**, union operators **107 → 52** (−55).

## Benchmark — honest characterization

**Deterministic:** the operator drop is `Σ max(0, kᵢ−2)` per union. On the airline policy: **107 → 52 union operators (−55, ~6% of the dataflow)**.

**Wall-clock scales with union density**, so it splits into two regimes:

| workload | metric | result |
| --- | --- | --- |
| **airline** real tau2, `-w1` (unions = ~6% of ops) | construction / commit | **within build-layout noise** — 3 release builds per variant: construction baseline mean 3406 µs vs fusion 3378 µs (−0.8%), inside the ~4.5% spread seen across builds of a *single* variant. Neutral / no regression, not a measurable speedup. |
| **union-dominated microbench** (single k-way union, k=200/400, batch) | construction | **−26% … −30%** (~2.3 µs per eliminated concat operator) |

So this is a **deterministic, parity-preserving operator-count reduction** and the idiomatic union spelling: a clear wall-clock win on union-dense programs, and neutral (no regression) on union-sparse ones like the airline policy.

🤖 Prototyped, benchmarked and validated with GitHub Copilot CLI.

### DIFF
--- a/flowlog-build/src/codegen/flow/non_recursive.rs
+++ b/flowlog-build/src/codegen/flow/non_recursive.rs
@@ -81,19 +81,21 @@ impl CodeGen {
                 .map(|fp| format_ident!("t_{}", fp))
                 .collect();
 
-            // Union the per-head collections left-to-right.
+            // Union the per-head collections.
             let head = &outs[0];
-            let mut expr: TokenStream = quote! { #head.clone() };
-            for t in &outs[1..] {
-                expr = quote! { #expr.concat(#t.clone()) };
-            }
+            let tail = &outs[1..];
 
             // Fold into the existing binding rather than shadowing it.
             let already_bound = bound_fps.contains(idb_fp);
             let (concat_expr, concat_count) = if already_bound {
-                (quote! { #output.concat(#expr) }, outs.len() as u32)
+                (quote! { #output.concatenate([ #( #outs.clone() ),* ]) }, 1)
+            } else if tail.is_empty() {
+                (quote! { #head.clone() }, 0)
             } else {
-                (expr, outs.len() as u32 - 1)
+                (
+                    quote! { #head.clone().concatenate([ #( #tail.clone() ),* ]) },
+                    1,
+                )
             };
 
             with_profiler(profiler, |profiler| {
--- a/flowlog-build/src/codegen/flow/recursive.rs
+++ b/flowlog-build/src/codegen/flow/recursive.rs
@@ -261,9 +261,11 @@ impl CodeGen {
                 ))
             })?;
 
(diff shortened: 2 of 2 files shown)


HarukiMoriarty (approved):
LGTM.

PR #208, opened 2026-07-05
fix(codegen): deterministic `ord` via single-thread fact interning

hdz284:
## Summary

`ord(x)` compiles to `x`'s string-interning key (a `lasso` `Spur`, handed out in
insertion order). The binary loader reads each fact file as parallel byte-ranges,
so under `-w N>1` workers intern their slices concurrently — every `ord` value,
and any `min(ord(..))` representative built on it (e.g. the DOOP port's
`HeapAllocation_Merge`), becomes **nondeterministic** across worker counts and
even run-to-run. Only `-w 1` was stable.

**Fix:** when the program uses `ord`, load facts on **worker 0 alone**, in the
same order the single-thread run uses; evaluation still runs on every worker.
So `-w N` interning order == `-w 1` == Souffle, and `ord` (with every
`min(ord(..))`) is deterministic and byte-exact again. `ord`'s value is
unchanged — **no fixtures or reference outputs need re-baselining.**

## Why gate on `ord` (addresses the review on #208)

`ord` is the **only** construct that can observe interning order, so serialising
the load for *every* `--str-intern` program was unnecessary:

- equality on interned strings is a bijection (`Spur` eq == content eq);
- ordered string comparison emits `resolve(l) < resolve(r)` — resolves to content;
- `min`/`max` semirings are numeric-only, so a string extremum must go through `ord`.

Hence `deterministic_load = str_intern_enabled() && program.uses_ord()`. Every
`--str-intern` program that never calls `ord` keeps the fully parallel loader at
**no cost**. New `Program::uses_ord()` also walks `loop`/`fixpoint` blocks (a gap
the typechecker's own `ord` check has via `Segment::as_rules`).

## Validation

- `cargo test --release --workspace` — **328 passed / 0 failed** (incl. 5 new
  `uses_ord` unit tests). `cargo fmt --check` clean. Rebased on `main-next`.
- **Full DOOP standalone re-verification** — 19 families, FlowLog `-w32` vs
  Souffle `-j32` (compiled *and* run threaded), luindex, each run **twice**:
  **19/19 byte-exact to Souffle AND run-to-run deterministic**, every count ==
  the `-w1` reference. The worst pre-fix family
  (`1-object-1-type-sensitive+heap`, which drifted ~146 K tuples run-to-run) is
  now MATCH.

| | before fix (`-w32`) | after fix (`-w32`) |
|---|---|---|
| vs Souffle, 19 families | 19/19 DIFF | **19/19 MATCH** |
| run A vs run B | nondeterministic | **byte-identical** |
| `ord` value / references | — | unchanged |

Overhead: nil for `ord`-free programs; a bounded ~2 s worker-0 load for `ord`
programs (FlowLog `-w32` timings/RSS match the pre-fix build, e.g.
context-insensitive 19.1 s / 4.5 GB). Full runtime + peak-RSS table available.

Updated from the previous single-thread-interning commit with the requested
`ord`-gate refinement, and rebased onto `main-next`.

### DIFF
--- a/flowlog-compiler/src/io/input.rs
+++ b/flowlog-compiler/src/io/input.rs
@@ -48,7 +48,12 @@ impl Compiler {
         let has_inline_facts = !self.program.facts().is_empty();
         let needs_preload = has_file_backed_edbs || has_inline_facts;
 
-        let maybe_peers = if has_file_backed_edbs {
+        // We want a deterministic load whenever the program uses `ord`,
+        // so its value depends on interning order, deterministic across
+        // different runs; evaluation still runs on every worker.
+        let deterministic_load = self.config.serialize_load();
+
+        let maybe_peers = if has_file_backed_edbs && !deterministic_load {
             quote! { let peers = worker.peers(); }
         } else {
             quote! {}
@@ -70,9 +75,21 @@ impl Compiler {
                             .into_owned()
                     })
                     .unwrap_or_else(|| file_name);
-                quote! {
-                    rels.get_mut(#rel_name).unwrap()
-                        .apply_file(std::path::Path::new(#path), SEMIRING_ONE, peers, index);
+                if deterministic_load {
+                    // Worker 0 alone reads the whole file (`peers = 1, index = 0`);
+                    // other workers skip loading. Interning order thus matches `-w 1`.
+                    quote! {
+                        if index == 0 {
+                            rels.get_mut(#rel_name).unwrap()
+                                .apply_file(std::path::Path::new(#path), SEMIRING_ONE, 1, 0);
+                        }
+                    }
+                } else {
+                    // Each worker ingests its own ~1/N byte-range slice in parallel.
+                    quote! {
+                        rels.get_mut(#rel_name).unwrap()
+                            .apply_file(std::path::Path::new(#path), SEMIRING_ONE, peers, index);
+                    }
                 }
             })
             .collect();
--- a/flowlog-build/src/build/pipeline.rs
+++ b/flowlog-build/src/build/pipeline.rs
@@ -52,9 +52,9 @@ impl Pipeline {
             ))
         })?;
 
-        let config = build_config(builder, program_str);
+        let mut config = build_config(builder, program_str);
         let mut program = parse(&config, &builder.include_dirs, sm)?;
-        check_program(&mut program, &config)?;
+        check_program(&mut program, &mut config)?;
         // Constant-fold after type checking (literals pinned, casts stripped)
         // and before planning, so the catalog and dataflow never see constant
         // sub-expressions.
@@ -112,5 +112,6 @@ fn build_config(builder: &Builder, program: &str) -> Config {
             .map(|p| p.to_string_lossy().into_owned())
             .collect(),
         output_to_stdout: false,
+        serialize_load: false,
     }
 }
--- a/flowlog-build/src/planner/program_planner.rs
+++ b/flowlog-build/src/planner/program_planner.rs
@@ -126,7 +126,7 @@ mod tests {
         let mut sm = SourceMap::new();
         let mut program =
             Program::parse(&tmp.path().to_string_lossy(), false, &[], &mut sm).expect("parse");
-        check_program(&mut program, &Config::default()).expect("typecheck");
+        check_program(&mut program, &mut Config::default()).expect("typecheck");
         ProgramPlanner::from_program(&Config::default(), &program, &mut None).expect("plan")
     }
 
--- a/flowlog-build/tests/typechecker_errors.rs
+++ b/flowlog-build/tests/typechecker_errors.rs
@@ -13,7 +13,7 @@ fn typecheck(name: &str) -> (Result<(), TypeCheckError>, SourceMap) {
     let mut sm = SourceMap::new();
     let mut program = Program::parse(&fixture("typechecker", name), false, &[], &mut sm)
         .expect("fixture should parse cleanly");
-    (check_program(&mut program, &Config::default()), sm)
+    (check_program(&mut program, &mut Config::default()), sm)
 }
 
 #[test]
--- a/flowlog-common/src/config.rs
+++ b/flowlog-common/src/config.rs
@@ -54,6 +54,10 @@ pub struct Config {
     /// Whether `.output` relations drain to stdout (`-D -`) rather than files.
     /// Derived by the CLI from `--output-dir`; always `false` in library mode.
     pub output_to_stdout: bool,
+    /// When set, fact strings are interned serially rather than in parallel, so
+    /// interning order — and therefore `ord(_)` values — is deterministic across
+    /// worker counts (`-w N` matches `-w 1`).
+    pub serialize_load: bool,
 }
 
 impl Config {
@@ -129,6 +133,11 @@ impl Config {
     /// Whether `.output` relations drain to stdout rather than files.
     pub fn output_to_stdout(&self) -> bool {
         self.output_to_stdout
+    }
+
+    /// Whether fact-string interning must be serial for deterministic `ord(_)`.
+    pub fn serialize_load(&self) -> bool {
+        self.serialize_load
     }
 }
 
--- a/flowlog-compiler/src/cli.rs
+++ b/flowlog-compiler/src/cli.rs
@@ -86,6 +86,7 @@ impl Cli {
             udf_file: self.udf_file.clone(),
             include_dirs: self.include_dirs.clone(),
             output_to_stdout: self.output_dir.as_deref() == Some("-"),
+            serialize_load: false,
         }
     }
 
--- a/flowlog-compiler/src/main.rs
+++ b/flowlog-compiler/src/main.rs
@@ -23,7 +23,7 @@ fn main() {
         .init();
 
     let cli = Cli::parse();
-    let config = cli.to_config();
+    let mut config = cli.to_config();
     let options = cli.to_compile_options();
 
     // Parse the source into an AST.
@@ -37,7 +37,7 @@ fn main() {
     .unwrap_or_else(|err| emit_and_exit(err, &sm));
 
     // Type-check the program.
-    check_program(&mut program, &config).unwrap_or_else(|err| emit_and_exit(err, &sm));
+    check_program(&mut program, &mut config).unwrap_or_else(|err| emit_and_exit(err, &sm));
 
     // Constant-fold after type checking, before planning.
     fold_constants(&mut program);
--- a/flowlog-typechecker/src/lib.rs
+++ b/flowlog-typechecker/src/lib.rs
@@ -78,11 +78,11 @@ use crate::env::PrimitiveEnv;
 /// Runs two passes: Pass 1 ([`primitive`]) checks primitive `DataType`s and
 /// pins literals; Pass 2 ([`subtype`]) enforces subtype rules and lowers
 /// `as()` casts. `config` is consulted for config-gated builtins.
-pub fn check_program(program: &mut Program, config: &Config) -> Result<(), TypeCheckError> {
+pub fn check_program(program: &mut Program, config: &mut Config) -> Result<(), TypeCheckError> {
     let env = PrimitiveEnv::from_program(program);
 
     primitive::rule::check_and_pin_rules(program, &env)?;
-    primitive::check_builtin_config_requirements(program, config)?;
+    primitive::builtin::check_ord(program, config)?;
     primitive::fact::check_and_pin_facts(program.facts_mut(), &env)?;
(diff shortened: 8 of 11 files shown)


flowlog-compiler/src/io/input.rs
```diff
@@ -70,9 +85,21 @@ impl Compiler {
...
+                    // Worker 0 alone reads the whole file (`peers = 1, index = 0`);
+                    // other workers skip loading. Interning order thus matches `-w 1`.
+                    quote! {
+                        if index == 0 {
+                            rels.get_mut(#rel_name).unwrap()
+                                .apply_file(std::path::Path::new(#path), SEMIRING_ONE, 1, 0);
+                        }
+                    }
+                } else {
+                    // Each worker ingests its own ~1/N byte-range slice in parallel.
```
HarukiMoriarty, line 88: Only when `ord` builtin function is used we need the deterministic loading. `ord` is default require to enable `str-intern`.

HarukiMoriarty (changes requested):

flowlog-compiler/src/io/input.rs
hdz284: Done — the deterministic load is now gated on actual `ord` use: `deterministic_load = str_intern_enabled() && program.uses_ord()`. `--str-intern` programs that never call `ord` keep the fully parallel loader. `Program::uses_ord()` also walks `loop`/`fixpoint` blocks. Rebased onto main-next; `cargo test` 328/0 and the full 19-family DOOP suite is byte-exact + deterministic at -w32.

HarukiMoriarty (changes requested):
Fix is correct and the determinism write-up is great. One structural ask before merge: the new `ord`-detection (`flowlog-parser/src/program/builtins.rs` + `mod builtins;` + its 5 tests) re-implements a walk `flowlog-typechecker::check_builtin_config_requirements` (`primitive/mod.rs:38`) already does — same rule/head/compare/factor traversal, already branches on `BuiltinOperator::Ord`, already has `config` in hand. Let's make "needs deterministic load" a **side effect of that existing check**, carried on the unified `Config`, and drop the standalone walker.

### Guidance

**1. `flowlog-common::Config` — add a derived flag** (same category as the existing derived `output_to_stdout`):

```rust
pub struct Config {
    // ...
    /// Set by the typechecker when the program uses `ord(_)`. Forces the
    /// binary-mode loader to intern on worker 0 alone so `ord` keys match
    /// `-w 1`. Derived from the program, not user-supplied.
    pub serialize_load: bool,
}

// getter, next to str_intern_enabled():
pub fn serialize_load(&self) -> bool { self.serialize_load }
```

The 4 exhaustive `Config { .. }` literals each need `serialize_load: false`: `cli.rs:80`, `pipeline.rs:99`, and the two test literals (`dedup.rs:96`, `planner_errors.rs:20`).

**2. `check_program` — take `&mut Config`** (`lib.rs:81`) and fold detection into the one walk. Rewriting `check_builtin_config_requirements` as a pure "first `ord` span" finder + a decision at the top keeps the nested fns capture-free and does both jobs:

```rust
pub(crate) fn resolve_ord_load(
    program: &Program,
    config: &mut Config,
) -> Result<(), TypeCheckError> {
    fn arith(a: &Arithmetic) -> Option<Span> {
        factor(a.init()).or_else(|| a.rest().iter().find_map(|(_, f)| factor(f)))
    }
    fn factor(f: &Factor) -> Option<Span> {
        match f {
            Factor::Var(_) | Factor::Const(_) => None,
            Factor::FnCall(fc) => fc.args().iter().find_map(arith),
            Factor::Builtin(bc) if bc.op() == BuiltinOperator::Ord => Some(bc.span()),
            Factor::Builtin(bc) => bc.args().iter().find_map(arith),
            Factor::Cast(c) => factor(c.inner()),
            Factor::Group(a) => arith(a),
            Factor::Tuple(t) => t.exprs().find_map(arith),
            Factor::TupleProj { tuple, .. } => arith(tuple),
        }
    }

    // NOTE: chain `as_loop()` — the current loop only visits `as_rules()`
    // (line 69), which skips `loop`/`fixpoint` blocks.
    let ord_span = program.segments().iter().find_map(|seg| {
        let plain = seg.as_rules().iter();
        let looped = seg.as_loop().into_iter().flat_map(|b| b.rules().iter());
        plain.chain(looped).find_map(|rule| {
            let body = rule.rhs().iter().find_map(|p| match p {
                Predicate::PositiveAtom(_) | Predicate::NegativeAtom(_) => None,
                Predicate::Compare(c) => arith(c.left()).or_else(|| arith(c.right())),
            });
            body.or_else(|| {
                rule.head().head_arguments().iter().find_map(|h| match h {
                    HeadArg::Var(_) => None,
                    HeadArg::Arith(a) => arith(a),
                    HeadArg::Aggregation(agg) => arith(agg.arithmetic()),
                })
            })
        })
    });

    if let Some(span) = ord_span {
        if !config.str_intern_enabled() {
            return Err(TypeCheckError::OrdRequiresStrIntern { span });
        }
        config.serialize_load = true; // the side effect
    }
    Ok(())
}
```

Two things this fixes vs. the current function: (a) it no longer early-returns when `str_intern` is on (`primitive/mod.rs:42-43`) — that early-return is why the `ord` + `--str-intern` case would otherwise never be walked; (b) it now walks loop/fixpoint rules, which also closes a latent gap where `ord` inside a `fixpoint` with `--str-intern` off currently escapes `OrdRequiresStrIntern`. Update the call in `check_program` (`lib.rs:85`) to the new name.

**3. `io/input.rs:67` — consume the flag**, drop the `Program::uses_ord()` call:

```rust
let deterministic_load = self.config.serialize_load();
```

`str_intern` is implied (the flag is only ever set under it), so both the `&&` and `uses_ord()` disappear.

**4. Delete** `flowlog-parser/src/program/builtins.rs` and the `mod builtins;` line in `program/mod.rs`. Move the essential cases — positive, comparison, `min(ord())` aggregation, and the fixpoint-nested regression — into a `flowlog-typechecker` test asserting the side effect:

```rust
let mut cfg = Config { str_intern: true, ..Default::default() };
check_program(&mut prog, &mut cfg).unwrap();
assert!(cfg.serialize_load);
```

**5.** Point the 5 `check_program` call sites at `&mut`: `main.rs:40`, `pipeline.rs:57`, `common.rs:32`, `typechecker_errors.rs:16`, `program_planner.rs:129` (the `&Config::default()` ones become `&mut Config::default()`).

One nuance worth a code comment: the flag is set at typecheck, before const-fold/DCE, so a dead `ord`-bearing rule that's later eliminated leaves it conservatively `true` — an unnecessary serial load, never a wrong result.

HarukiMoriarty:
Minor / non-blocking — just to scope the guarantee: worker-0 load pins interning order for strings that come from **input files / inline facts**, but `ord` of a string first created *during evaluation* (a rule-head constant like `Tag("x")`, or a `cat`/`substr`/`to_string`/UDF result) is still interned in parallel and stays order-dependent under `-w N`.

That's fine for the DOOP `min(ord(v))` usage (all `v` are loaded symbols), and Soufflé doesn't guarantee it under `-j N` either — so nothing to fix. Might just be worth narrowing the `input.rs` comment's *"byte-for-byte identical to `-w 1`"* to `ord` of loaded/inline symbols.

HarukiMoriarty:
@hdz284 quick reminder to sign off the commits — `76c096d` (`docs(runtime): …`) is missing its `Signed-off-by:` line (`82b1049` already has it). Mind adding the DCO sign-off before merge? e.g. `git rebase --signoff HEAD~2 && git push -f`.

hdz284:
Done — folded the ord-detection into the single config-requirements walk as suggested (2f70c22).

- **`Config::serialize_load`** added as a derived flag next to `output_to_stdout` (+ getter); the two exhaustive literals (`cli.rs`, `pipeline.rs`) set `false`, the rest inherit via `..Default::default()`.
- **`check_builtin_config_requirements` → `resolve_ord_load(program, &mut Config)`**: a first-`ord`-span finder that errors `OrdRequiresStrIntern` when str-intern is off, else sets `serialize_load`. `check_program` now takes `&mut Config`; all 5 call sites updated.
- **`input.rs`** consumes `config.serialize_load()`; the `str_intern && uses_ord()` gate is gone (str-intern is implied since the flag is only set under it).
- **Deleted** `flowlog-parser/src/program/builtins.rs` + `mod builtins;` + its 5 parser tests; re-homed coverage as `flowlog-typechecker/tests/ord_load.rs` asserting the `serialize_load` side effect (positive/comparison/`min(ord())`/fixpoint) plus the two error cases.

Both latent bugs you flagged are fixed and regression-tested: the walk no longer early-returns under str-intern, and it chains `as_loop()` so `ord` inside a `fixpoint` is caught (new `ord_inside_fixpoint_without_str_intern_is_rejected` test). Added the code comment on the pre-fold/DCE conservatism.

**Verification:** 363/0 workspace tests. Rebuilt the release compiler and re-ran the DOOP standalone check on luindex — `context-insensitive` and `1-type-sensitive` recompiled, both **deterministic** (`-w32` run1 == run2) and **byte-exact vs Soufflé** (0 tuples either way), i.e. the fix's behaviour is unchanged by the refactor.

hdz284:
### Re-benchmarked all 19 DOOP families with the refactored compiler ✅

Rebuilt the release compiler at 2f70c22 and re-ran the full DOOP standalone suite on `luindex` (`-w32` vs Soufflé `-j32`). Every family is **deterministic** (`-w32` run1 == run2) and **byte-exact vs Soufflé** — identical to the pre-refactor results, confirming the ord-load consolidation is behaviour-preserving:

| | families |
|---|---|
| deterministic (`-w32` run1==run2) | **19 / 19** |
| byte-exact vs Soufflé | **19 / 19** |

Row counts match the reference exactly (e.g. `1-type-sensitive` 9 206 109, the three object+type analyses 8 719 851 / 8 719 851 / 9 788 238). `0` tuples unique to either engine on all 19. Combined with the 363/0 workspace tests, the `serialize_load` side effect is wired end-to-end (typecheck → `Config` → codegen loader) with no behavioural change.

hdz284:
## [w4] Independent verification of the ord-load refactor ✅

I'm an AI agent (tagged **[w4]**). I picked up @HarukiMoriarty's review independently and, before seeing `2f70c22`, implemented the same change from scratch — and **converged on essentially the identical design**: a single `resolve_ord_load` walk, a derived `Config::serialize_load` (getter alongside `output_to_stdout`), `check_program(&mut Config)`, the `as_loop()`-chained traversal, `builtins.rs` deleted, and a `flowlog-typechecker/tests/ord_load.rs` asserting the side effect. Two independent implementations landing on the same shape is a good signal the reviewer's suggested structure is the natural one.

### Independent byte-exact reproduction
I built the refactored release compiler myself and re-ran a cross-section of the DOOP standalone suite on `luindex` (`-w32` FlowLog vs `-j32` Soufflé 2.5, byte-exact after canonicalize + `LC_ALL=C sort`):

| Family | rows (both engines) | only-FL | only-SF | verdict |
|---|--:|--:|--:|:--:|
| context-insensitive | 5,396,536 | 0 | 0 | ✅ |
| 1-object-sensitive | 15,093,978 | 0 | 0 | ✅ |
| 1-type-sensitive+heap | 11,736,906 | 0 | 0 | ✅ |
| 1-call-site-sensitive+heap | 17,573,686 | 0 | 0 | ✅ |
| 1-object-1-type-sensitive+heap | 8,719,851 | 0 | 0 | ✅ |

The **type-sensitive** and **type-object** families are the meaningful ones here: they were `-w32` *nondeterministic* before this PR, and are now byte-exact vs Soufflé (`1-object-1-type-sensitive+heap` = 8,719,851 matches its known `-w1` value). This corroborates the author's 19/19 result. I also confirmed via codegen that an `ord` program emits the worker-0 serial loader (`if index == 0 { apply_file(.., 1, 0) }`) while a non-`ord` `--str-intern` program keeps the parallel loader — so the `serialize_load` flag is wired typecheck → `Config` → codegen with no behavioural change. `cargo clippy --all-targets -- -D warnings` is clean.

### One open item before merge (DCO)
`76c096d` and `2f70c22` are still missing their `Signed-off-by` line — the refactor was addressed, but the sign-off from @HarukiMoriarty's earlier note wasn't. The DCO check will block merge until both commits are signed off (the earlier review comment has the exact `rebase --signoff` incantation).

Non-blocking nit: the `io/input.rs` "byte-for-byte identical to `-w 1`" comment could still be narrowed to `ord` of **loaded/inline** symbols (strings first created during evaluation are still interned in parallel, as Soufflé is under `-j N`) — matching @HarukiMoriarty's earlier scoping note.

HarukiMoriarty (approved):
LGTM.

PR #218, opened 2026-07-09
[w3] Fuse spanning equalities into join keys (11–16× on DOOP type-object analyses; byte-exact; beats Souffle)

hdz284:
## TL;DR

When two body atoms share no bare variable, the planner emitted an **empty-key cross-join + post-join filter** — an `O(|L|·|R|)` blow-up. This PR **fuses a spanning equality `L == R` into the join key** (a real hash join), where `L` is grounded wholly in one atom and `R` in the other. It covers tuple-projection keys (the DOOP `hctx.0 == hctxValue`) and arithmetic keys (`x + 1 == y + 2`).

On the DOOP standalone suite this makes the three type-object analyses **11–16× faster**, is **byte-exact on all 19 families**, and makes **FlowLog beat Souffle on every family** (previously it lost these three by 6.5–7×).

## The problem

FlowLog keys a join only on variables that appear **by name** in both atoms. A join expressed as a cross-atom equality — e.g. DOOP's `hctx = (hctxValue,), Value_DeclaringType(hctxValue, type)`, which desugars to `Value_DeclaringType(hctx.0, type)` — has no shared name, so it became a cross product with `hctx.0 == hctxValue` as a residual filter. On `luindex` that single rule dominated runtime (single-worker cross product; the other 31 workers idle).

## The fix

- **`catalog::equijoin_keys_for_pair`** — resolves each *spanning* `=` comparison (empty superset) to a computed `(lhs_key, rhs_key)` `ArithmeticPos` pair; `ArithmeticPos::is_plain_var` separates column selection from a computed key.
- **`planner::apply_join`** — keys the join on the fused equality instead of producing an empty-key cross product.
- **layout propagation (`fuse.rs` + `transformation/info.rs`)** — `refactor_output_key_value_layout` carries the *computed* key through to the input-arrangement reshape so it **materializes** the derived value (`hctx.0`) instead of collapsing to the base column (`hctx`). This is the crux: FlowLog's arrangements are column-selection based, so the projection has to be materialized as the key.

**Safety / generality.** Fresh-variable equalities (`z = g(x)`) are desugared to bindings *before* planning, so only both-sides-grounded equalities ever reach the fusion — exactly the ones safe to key on. The generated join closure keeps the equality as a redundant residual, so correctness holds for any shape. The change reuses existing machinery (comparison/superset analysis, the per-join-rebuilt variable-presence map, variable-lifetime pruning), so it generalizes to multi-atom join orders, composite keys, and keys whose base variable is not in the head. Positive-join only; antijoins are untouched.

Verified with a new planner unit test + 6 synthetic end-to-end tests (tuple projection; arithmetic on both sides; fresh-var *not* fused; base-var-not-in-head; non-adjacent 3-atom join; composite multi-key). **All 301 workspace tests pass.**

## Benchmark configuration

| | |
|---|---|
| **Machine** | AMD EPYC 7763, 64 cores, 503 GB RAM |
| **Programs / data** | DOOP standalone (`flowlog-rs/doop-flowlog`, `verify/standalone`), 19 context families |
| **Primary dataset** | `luindex` facts (130 files, 713 MB); plus `lusearch`, `avrora`, `sunflow`, `pmd` (0.7–1.3 GB) for the cross-dataset runs |
| **FlowLog** | base `nemo/tuple` (32cbc00) + this PR; `--mode datalog-batch`, `--str-intern`, `-w 32` |
| **Souffle** | 2.5, compiled `souffle -c -j 32`, run `-j 32` |
| **Correctness oracle** | byte-exact after canonicalization (bracket/space normalize + `LC_ALL=C sort`), compared by `sha256` |

Notes: FlowLog's generated crate was built at `opt-level 1` (dependencies incl. Differential Dataflow / timely stay at `opt-level 3`) purely to keep the ~29k-line generated function's LLVM compile tractable — a **conservative handicap** for FlowLog in the vs-Souffle numbers. Determinism was validated (run1 == run2) with the ord-determinism fix in place.

## Correctness (FlowLog vs Souffle)

- **All 19 DOOP families byte-exact** vs the Souffle/baseline reference on `luindex`. **Zero regressions.**
- **Cross-dataset:** the three fixed programs across **4 more datasets** (`lusearch`, `avrora`, `sunflow`, `pmd`) are **byte-exact vs Souffle** — **15/15** combos including `luindex`. Two non-target programs on `sunflow` (`context-insensitive`, `1-object-sensitive+heap`) are also byte-exact — **no regression** where the fusion doesn't fire.

## Performance

**The three type-object analyses (`luindex`, `-w32` / `-j32`):**

| Family | Baseline FL | **Fused FL** | Souffle | Speedup vs base | Fused vs Souffle |
|---|---|---|---|---|---|
| 1-object-1-type-sensitive+heap | 708 s | **44.6 s** | 110 s | **15.9×** | **2.5× faster** |
| 2-type-object-sensitive+heap | 724 s | **44.3 s** | 106 s | **16.3×** | **2.4× faster** |
| 2-type-object-sensitive+2-heap | 958 s | **87.0 s** | 138 s | **11.0×** | **1.6× faster** |

**The other 16 families:** all **1.0–1.6× faster** than baseline (never slower) and all beat Souffle. Across the full suite FlowLog now wins **1.4×–5.3×** vs Souffle.

**Cross-dataset (fused FlowLog vs Souffle, `-w32` / `-j32`; every cell byte-exact):**

| Dataset | 1obj-1type+heap | 2type-obj+heap | 2type-obj+2heap |
|---|---|---|---|
| luindex | 44.6 s (**2.5×**) | 44.3 s (**2.4×**) | 87.0 s (**1.6×**) |
| lusearch | 48.5 s (**2.0×**) | 40.4 s (**2.4×**) | 74.9 s (**1.7×**) |
| avrora | 46.8 s (**1.3×**) | 35.3 s (**1.6×**) | 66.3 s (**1.2×**) |
| sunflow | 51.6 s (**1.7×**) | 40.6 s (**2.1×**) | 79.7 s (**1.4×**) |
| pmd | 63.4 s (**1.7×**) | 46.7 s (**2.3×**) | 81.6 s (**1.5×**) |

Speedup vs Souffle; **15/15 byte-exact**, FlowLog **1.2–2.5×** across all. `jython` (1.8 GB, reflection-heavy) was **excluded**: the deep context-sensitive configs are intractable on **both** engines (FlowLog ~450/503 GB RAM, Souffle >4 h, neither wrote a `VarPointsTo` — a dataset limit, not a fusion issue).

## Implications

- Removes a pathological cross-join class from FlowLog's planner: any join expressible as a cross-atom equality (tuple field, arithmetic, builtin) is now a hash join.
- With this change FlowLog is **faster than Souffle on all 19 DOOP standalone families** while remaining **byte-exact** — including the three type-object analyses it previously lost by 6.5–7×.
- Peak RSS is higher than Souffle (arrangement-heavy), unchanged by this PR.

## Scope / reproduction

This PR is the **planner change only** (5 files, `flowlog-build`). Reproducing the DOOP run on `nemo/tuple` additionally needs two environment patches kept out of this focused PR: tuple-typed EDB ingest, and the ord-determinism port (#208); plus the generated-crate `opt-level` override above. The fusion itself is independent of those and builds/tests clean on stock `nemo/tuple`.

### DIFF
--- a/flowlog-build/src/catalog/arithmetic.rs
+++ b/flowlog-build/src/catalog/arithmetic.rs
@@ -256,6 +256,16 @@ impl ArithmeticPos {
         &self.rest
     }
 
+    /// Whether this position is a bare variable reference (no arithmetic tail and
+    /// no derivation on the initial factor). Such a position is a plain column of
+    /// its atom and can be selected by index; anything else (a tuple projection,
+    /// builtin, UDF call, constant, …) is a *computed* value that must be
+    /// materialized rather than selected.
+    #[inline]
+    pub(crate) fn is_plain_var(&self) -> bool {
+        self.rest.is_empty() && self.init.as_var_signature().is_some()
+    }
+
     /// Returns a vector of all variable signatures referenced in this expression, in order.
     pub(crate) fn signatures(&self) -> Vec<&AtomArgumentSignature> {
         let mut sigs = self.init.signatures();
--- a/flowlog-build/src/catalog/rule.rs
+++ b/flowlog-build/src/catalog/rule.rs
@@ -8,7 +8,7 @@ use crate::catalog::{
     FnCallPredicatePos,
 };
 use crate::common::{SECTION_BAR, SUBSECTION_BAR};
-use crate::parser::{ComparisonExpr, FlowLogRule, FnCall, HeadArg, Predicate};
+use crate::parser::{Arithmetic, ComparisonExpr, ComparisonOperator, FlowLogRule, FnCall, HeadArg, Predicate};
 use tracing::debug;
 
 // Implementation modules
@@ -411,6 +411,62 @@ impl Catalog {
             &left_var_signatures,
             &right_var_signatures,
         ))
+    }
+
+    /// Equi-join fusion: for each equality comparison `L = R` whose left side is
+    /// fully grounded in positive atom `lhs_idx` and right side in `rhs_idx` (or
+    /// vice-versa), return the `(lhs_key, rhs_key)` computed-key pair so the
+    /// caller can key the join on `L == R` (a hash join) instead of a cross-join
+    /// plus a post-join filter.
+    ///
+    /// Only `=` comparisons qualify, and only *spanning* ones (empty superset —
+    /// not evaluable against any single atom yet). Fresh-variable equalities
+    /// (`z = g(x)`) are desugared to bindings upstream, so `comparison_predicates`
+    /// only ever holds equalities between two already-grounded expressions, which
+    /// is exactly what is safe to fuse into a join key.
+    pub(crate) fn equijoin_keys_for_pair(
+        &self,
+        lhs_idx: usize,
+        rhs_idx: usize,
+    ) -> Vec<(ArithmeticPos, ArithmeticPos)> {
+        // Resolve every variable of `arith` to its signature in positive atom
+        // `atom_idx`; returns None if any variable is absent from that atom (so
+        // the side does not live wholly inside `atom_idx`).
+        let resolve = |arith: &Arithmetic, atom_idx: usize| -> Option<ArithmeticPos> {
+            let mut sigs = Vec::new();
+            for v in arith.vars() {
+                let sig = self
+                    .argument_presence_in_positive_atom_map
+                    .get(v)
+                    .and_then(|row| row.get(atom_idx).copied().flatten())?;
+                sigs.push(sig);
+            }
+            Some(ArithmeticPos::from_arithmetic(arith, &sigs))
+        };
+
+        let mut out = Vec::new();
+        for (comp_id, comp) in self.comparison_predicates.iter().enumerate() {
+            if *comp.operator() != ComparisonOperator::Equal {
+                continue;
+            }
+            // Skip comparisons already coverable by a single atom (within-atom
+            // filters / already-joined) — only genuine cross-atom equalities.
+            if !self.comparison_supersets[comp_id].is_empty() {
+                continue;
+            }
+            if let (Some(l), Some(r)) =
+                (resolve(comp.left(), lhs_idx), resolve(comp.right(), rhs_idx))
+            {
+                out.push((l, r));
+            } else if let (Some(l), Some(r)) =
+                (resolve(comp.left(), rhs_idx), resolve(comp.right(), lhs_idx))
+            {
+                // left lives in the rhs atom, right in the lhs atom: swap so that
+                // the first element always keys `lhs_idx`.
+                out.push((r, l));
+            }
+        }
+        out
     }
 
     // === FnCall Predicates ===
--- a/flowlog-build/src/planner/rule_planner/core.rs
+++ b/flowlog-build/src/planner/rule_planner/core.rs
@@ -116,22 +116,44 @@ impl RulePlanner {
         )?;
 
         // Partition arguments into join keys and payload values
-        let (lhs_keys, lhs_vals, rhs_keys, rhs_vals) = Self::partition_shared_keys(
+        let (mut lhs_keys, lhs_vals, mut rhs_keys, rhs_vals) = Self::partition_shared_keys(
             catalog,
             left_atom_argument_signatures,
             right_atom_argument_signatures,
         );
-        fn labelled<'a>(
-            positions: &'a [ArithmeticPos],
-            catalog: &'a Catalog,
-        ) -> Vec<(&'a ArithmeticPos, &'a String)> {
+
+        // Equi-join fusion (see `Catalog::equijoin_keys_for_pair`): when the two
+        // atoms share no bare variable — which would otherwise force an empty-key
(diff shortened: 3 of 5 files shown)


hdz284:
## Independent reproduction on the DOOP standalone suite ✅

Verified this fusion end-to-end on `flowlog-rs/doop-flowlog` `verify/standalone` (luindex, `-w32`/`-j32`),
as a controlled A/B: this branch (`3a14b67`) vs its base `nemo/tuple` (`32cbc00`) — one commit apart,
with the two out-of-PR env patches (tuple-typed EDB handler + generated-crate `opt-level=1`) applied
**identically to both**, so the fusion is the only variable.

| Family | base FL | **fused FL** | Soufflé | **vs base** | **vs Soufflé** | (PR claim) |
|---|--:|--:|--:|--:|--:|--:|
| 1-object-1-type-sensitive+heap | 940 s | **60.9 s** | 136 s | **15.4×** | **2.24×** | 15.9× / 2.5× |
| 2-type-object-sensitive+heap | 918 s | **50.5 s** | 119 s | **18.2×** | **2.36×** | 16.3× / 2.4× |
| 2-type-object-sensitive+2-heap | 1174 s | **94.5 s** | 192 s | **12.4×** | **2.03×** | 11.0× / 1.6× |

**Correctness:** fused output is **byte-exact vs Soufflé** at `-w1` on all three
(8 719 851 / 8 719 851 / 9 788 238 rows, 0 tuples unique to either engine). Since `nemo/tuple` lacks
the ord fix (#208), the parallel path is nondeterministic — but base and w3 drift by the same margin,
so I checked correctness at `-w1` (deterministic); the fusion doesn't change the result set.

Numbers line up with the PR's (base times run ~1.3× high here due to a concurrent background load; the
ratios cancel it). **The 11–16× gains are real, and it turns FlowLog's only DOOP-standalone losses into
~2× wins over Soufflé.** Full write-up + scripts + env-patch diff:
flowlog-rs/doop-flowlog#2 (`verify/standalone/w3-fusion-check/`).

hdz284:
## [w4] Code review + independent baseline corroboration

I'm an AI agent (tagged **[w4]**). I read the whole diff and re-measured the *baseline* independently. To be upfront: I did **not** re-run the fused numbers (my environment is `main-next`+#208, not this PR's `nemo/tuple` base), so I'm corroborating the **problem** and the **baseline correctness**, and reviewing the **code** — not re-verifying the 11–16× itself.

### Independent corroboration of the motivation
On a *different* base (`main-next` + #208, **no fusion**), I ran `1-object-1-type-sensitive+heap` on `luindex`, `-w32` vs Soufflé `-j32`:

| | FlowLog (unfused) | Soufflé | FL vs SF |
|---|--:|--:|--:|
| 1-object-1-type-sensitive+heap | **631.2 s** | 108.4 s | **0.17× (5.8× slower)** |

Byte-exact (8,719,851 rows, 0 unique to either engine). This lines up with the PR's baseline (708 s / 110 s) and independently confirms two things: the empty-key cross-product blow-up is **real and base-independent** (it shows up on `main-next` too, not just `nemo/tuple`), and the unfused result is a **correct** baseline for the fusion to preserve. So the ~6× loss this PR targets is genuine.

### Code observations (from reading the diff)
The design reads cleanly and the safety argument holds up:

- **`equijoin_keys_for_pair`** correctly restricts to `=` with an **empty superset** (genuine cross-atom equalities). Fresh-var equalities are desugared upstream and single-atom-evaluable ones are filtered by the superset check, so only both-sides-grounded spanning equalities reach the key — the safe set. The residual equality kept in the join closure is a nice belt-and-suspenders.
- The subtle bit I was most worried about is handled: the group key in `collect_consumer_layout_indices` folds in a **per-slot computed signature**, so two consumers deriving *different* projections of the same base (`hctx.0` vs `hctx.1`) don't collapse onto one arrangement. Materializing the computed key in `refactor_output_key_value_layout` (plain vars stay index-selected) is the right split.
- Fusion only fires when `lhs_keys.is_empty()` — i.e. it *rescues* empty-key joins rather than adding keys to already-keyed ones. That's a reasonable, conservative scope; worth a one-line note in the code that a shared-var join with an extra spanning `=` keeps the `=` as a residual by design.

Minor / optional:
- `const ENABLE_EQUIJOIN_FUSION: bool = true;` reads like a kill-switch that's now permanently on — consider dropping it (or promoting it to a real `Config`/planner flag) once you're confident, so there's no dead `if true` branch long-term.
- Degenerate constant equalities (`1 = 2`, `1 = 1`) technically satisfy "empty superset"; they're semantically preserved (empty / tautological cross-product) and normally constant-folded away, so this is benign — just noting it since `resolve` returns `Some` for a var-free side.

Net: from a read-through + baseline-corroboration standpoint this looks solid and the test matrix (tuple projection, arithmetic both sides, fresh-var-not-fused, base-var-not-in-head, 3-atom non-adjacent, composite key) covers the shapes I'd have asked for. I just can't personally vouch for the fused wall-clock numbers.

HarukiMoriarty:
Rewritten in #219, close this.

PR #260, opened 2026-07-22
revert: drop workflow-based DCO, use the DCO app

HarukiMoriarty:
Removes the interim `dco.yml`. dcoapp/app #292 (merged 2026-07-22) adds merge_group support, so the app will cover the queue once its hosted instance deploys + the install accepts the new event. DCO is advisory in the meantime (already dropped from required checks), and the app still posts its check on PRs. Companion revert on `main` too. Re-require the app's DCO on both branches once it reports on `merge_group`.


HarukiMoriarty:
Keeping dco.yml as an advisory DCO check: the app was uninstalled, so this workflow is now the only thing posting a DCO status on PRs. DCO stays non-required (advisory) until dcoapp/app #292's merge_group support is deployed and the app reinstalled, then we re-require the app and delete this.

PR #261, opened 2026-07-22
revert: drop workflow-based DCO, use the DCO app (main copy)

HarukiMoriarty:
Companion to the `main-next` revert. Removes the registration copy of `dco.yml`; the DCO app covers DCO (advisory until its merge_group support deploys to the hosted instance).


HarukiMoriarty:
Companion close: keeping the workflow as an advisory DCO check now that the app is uninstalled.

PR #262, opened 2026-07-22
revert: use the DCO app (advisory) instead of the workflow

HarukiMoriarty:
Reverts the interim `dco.yml`. The DCO app is reinstalled and now posts the DCO check on PRs (advisory). DCO stays **non-required** until the app's `merge_group` support (dcoapp/app #292) deploys to the hosted instance and the install subscribes to `merge_group`; at that point we re-require it (force). Companion revert on `main`.


HarukiMoriarty:
Obsolete: dco.yml was removed from main-next by rewriting history (the add commit #258 was dropped), so there's nothing to revert. DCO is advisory via the reinstalled app.

PR #286, opened 2026-08-08
refactor!: drop extended execution modes and loop blocks

HarukiMoriarty:
Deletes the `ExtendBatch` / `ExtendInc` execution modes and the `loop` /
`fixpoint` block syntax they gated. Nobody uses them, they are a persistent
source of complexity, and the mode conflation they forced has already caused
one silent correctness bug.

Stacked on #284; review that first. Base retargets to `main-next` once it lands.

## `.iterative` was the load-bearing question

`iterative_directive` appears in the grammar only inside `fixpoint_block` and
`loop_block`, so deleting the blocks makes the entire replacement-semantics
path unreachable. That is most of the diff:

- `IterativeDirective`, `iterative_relations`, `iterative_recursive_relations`
  and `recursion_iterative_recursive_collections` through the stratifier and
  planner
- `Variable::new_from` — all recursion is accumulative `Variable::new`
- the mixed-loop leave arms in `build_leave_outputs`
- the loop gate: `ConditionPlan`, `build_iter_conditions`, `continue_stmt`,
  `stop_stmt`, `negated_weight_tokens`

`codegen/flow/recursive.rs` loses 439 lines, `stratifier/core.rs` 661.

## Also dead once the blocks go

- `writer_drops_sign`, guarded on `is_batch() && !is_datalog_batch()` — i.e.
  `ExtendBatch` alone
- five `StratifyError` variants and four `ParseError` variants
- the `in_loop` arm of constant folding, which only ever fired inside a block

The stratifier now runs one global SCC pass instead of coalescing runs between
block barriers.

## Naming

`ExecutionMode` collapses to `Batch` / `Inc`, and `is_batch()` /
`is_datalog_batch()` become one predicate — removing the "batch implies the
`Present` difference" conflation that was false for `ExtendBatch`. With one
dialect left the `Datalog` qualifier stops distinguishing anything, so the
generated engines drop it too.

## Consequence worth reviewing on its own

`Segment` had one variant left, so `Program` is now a flat rule list:
`Program::rules` returns a slice rather than allocating a `Vec<&FlowLogRule>`
per call, and `flush_rules`, `for_each_rule_mut` and the Plain-run coalescing
go with it. A `.init` splices its instantiated rules in at the index it held.

## Breaking

- `--mode` takes `batch` or `inc`
- `flowlog_common::ExecutionMode` variants renamed; `Config::is_extended` and
  `Config::is_datalog_batch` removed
- flowlog-parser no longer exports `LoopBlock`, `LoopCondition`,
  `LoopConnective`, `StopGroup`, `StopRelation`, `IterativeDirective`, or
  `Segment`
- generated crates expose `BatchEngine` / `IncrementalEngine`

## Verification

`cargo +nightly fmt --all --check`, `cargo clippy --workspace --all-targets
-- -D warnings`, `cargo test --workspace` (17 groups), and
`bash tests/fixtures/run_lib.sh` — **119/119**, down from 128 with the nine
`extend-batch` fixtures deleted. Fixture categories renamed `datalog-batch/`
to `batch/` and `datalog-inc/` to `inc/`; those 551 files are pure renames
(verified by blob SHA, not git's heuristic), so reviewing the 93 non-rename
files covers the whole change:

```
git diff --stat -M --diff-filter=DM
```

Compiler mode is unverified here for the pre-existing differential-dataflow
instance split the rest of this stack already crosses.

## Two open questions

1. `validate_forward_references` and `validate_recursive_strata` were
   reachable only through loop-block barriers. With one global SCC pass I
   believe neither can now fire — topological ordering rules out a forward
   reference, and a recursive SCC always has a head appearing as a body atom.
   I kept both as invariant checks and deleted only their barrier-dependent
   tests, so they are currently uncovered. Delete, convert to internal errors
   per `docs/dev/errors.md` rule 2, or leave?
2. `steps.rs`'s module doc said profiling was verified only for the Datalog
   modes because extended modes "may add operators this table does not know
   (loop conditions, UDF pipelines)". I dropped the sentence with the modes,
   but the UDF half may still apply — restore it?


HarukiMoriarty:
Resolving the two open questions:

**1. `validate_forward_references` / `validate_recursive_strata`: deleted** (c1d6eff). Neither can fire under the global SCC pass, by construction:

- *Forward references*: `DependencyGraph::from_rules` adds an edge from a rule to every rule deriving one of its body atoms, and `merge_strata` emits a component only once all its dependencies are emitted — so a definition is always in the same stratum (covered by the same-stratum `heads` check) or an earlier one. The earlier case is also airtight: a head consumed by a later stratum is by definition in the defining stratum's `later_body_atoms`, hence a leave relation, hence in every later stratum's `available_relations`.
- *Empty recursive strata*: a component is recursive only with >1 rule or a self-edge, and `merge_strata` never merges recursive components. Mutual reachability (or the self-edge) means some rule's body names another rule's head inside the component, so `heads ∩ body_atoms` is never empty.

Rather than keeping them as unreachable internal errors, they're gone: the invariants are tested directly in `scc.rs` (e.g. `merge_places_dependencies_before_consumers`), so the validators were re-verifying algorithm output, not guarding a reachable state. `Stratifier::from_program` is now infallible and the last two `StratifyError` variants leave with their module.

**2. The UDF half of the `steps.rs` caveat stays dropped.** A UDF call is generated as `udf::fn_name(arg.clone(), …)` — a plain expression inlined inside an existing map closure (`codegen/arg.rs`). It never introduces a dataflow operator, so UDFs cannot shift the operator counts this table predicts. The old sentence's two examples were loop conditions (deleted here) and UDF pipelines (never operator-visible), so nothing remains to warn about.

hdz284 (approved):

PR #302, opened 2026-08-12
perf: add low-overhead incremental engine options

hdz284:
## Summary

Add two independent, opt-in ways to reduce fixed transaction overhead in
library-mode `DatalogInc` engines:

- `Builder::inline_single_worker(true)` generates a one-worker engine that
  owns and steps its timely worker on the calling thread. It emits no worker
  thread, barrier protocol, shared transaction state, per-worker input slots,
  or shared output buffers.
- `Builder::trusted_set_inputs(true)` removes the set-normalization
  arrangement from caller-fed EDBs when the caller guarantees that committed
  deltas already describe valid set transitions.

Both options default to `false`. Existing generated engines are unchanged
unless a caller opts in.

## Why

For short-lived incremental workloads, useful dataflow work can be smaller
than the fixed cost of handing each commit to a worker thread and maintaining
an arrangement solely to clamp EDB multiplicities to 0/1.

The inline engine removes the synchronization path entirely rather than
special-casing it at runtime. Keeping it as a build-time engine shape also
keeps default engines `Send` and avoids generating both runtimes.

The trusted-input option is separate because it changes the caller contract.
It is useful for hosts that already track relation membership and emit only:

- an insertion while a fact is absent; and
- a retraction while that fact is resident.

The promise is unchecked. Relations initialized by compiler-owned `.fact`
rows retain normalization because those rows are outside the caller contract.
IDB and recursive deduplication are unchanged.

## API

```rust
flowlog_build::Builder::default()
    .mode(flowlog_build::ExecutionMode::DatalogInc)
    .inline_single_worker(true)
    .trusted_set_inputs(true)
    .compile(&["policy.dl"], &[] as &[&std::path::Path])?;
```

An inline build preserves `DatalogIncrementalEngine::new(workers)` so the same
call site compiles against either generated shape, but accepts only `new(1)`.
It is intentionally `!Send`. Inline builds are rejected outside `DatalogInc`
and when profiling is enabled.

## Benchmark

Measured against SASY `flowlog-bench-mini` at
`176743164cc2f8206fb90cdfcc60135c94a277d1`, with this branch based on
`main-next` at `5654726a7eb53e99d63e5d6d2df1679a02166608`.

Matrix:

- 11 policy workloads
- linear and fan-in graph shapes
- 8, 32, and 128 turns
- 66 cells total, one worker
- 15 repetitions at 8 turns, 10 at 32, and 5 at 128
- baseline run immediately before and after every cell
- variant order rotated between cells
- every FlowLog result parity-checked against Souffle

| Variant | Median FlowLog turn-time reduction |
|---|---:|
| Inline worker only | 6.4% |
| Trusted set inputs only | 6.0% |
| Combined | **11.7%** |

The combined build improved all 66 cells, ranging from 2.2% to 30.5%.
Median absolute baseline drift between the per-cell bookends was 0.9%
(p90 2.4%).

Representative combined median gains by workload:

| Workload | Gain |
|---|---:|
| OpenClaw | 29.1% |
| MLS | 25.3% |
| LLM content | 17.6% |
| Toxic flow | 14.5% |
| DLP guard | 12.6% |
| Airline | 6.9% |
| Retail | 5.9% |

This does **not** make FlowLog faster than Souffle in the small SASY cells.
FlowLog's median FlowLog/Souffle ratio improved from 17.5x to 12.1x, and the
number of FlowLog-winning cells remained 4/66 (the existing large fan-in
wins). Among cells where Souffle remained faster, the combined build closed a
median 13.6% of the absolute latency gap.

## Correctness and compatibility

- Default-off generated code remains unchanged.
- Inline and threaded engines share the same public transaction surface.
- Trusted inputs retain clamps on relations with inline `.fact` rows.
- Binary/compiler mode explicitly keeps both options off.
- Unsupported inline option combinations fail during `flowlog-build`
  compilation.
- The fixture runner supports `--inline`, `--trusted-inputs`, and category
  selection for reproducible option-specific coverage.

Validation:

- `cargo +nightly fmt --all --check`
- `cargo test --workspace` (707 passed)
- all 124 default library fixtures
- all 21 `DatalogInc` fixtures with inline + trusted inputs
- all 21 `DatalogInc` fixtures with trusted inputs and `WORKERS=2`


hdz284:
Follow-up performance audit after opening this PR:

I prototyped four additional candidates on top of the combined
`inline_single_worker + trusted_set_inputs` build and reran the same 66-cell
SASY matrix (11 workloads, two graph shapes, 8/32/128 turns, per-cell
bookended controls, parity-gated against Souffle).

| Candidate | Median additional FlowLog turn-time change | Decision |
|---|---:|---|
| Host-side epoch output netting instead of differential `.consolidate()` | **2.2% faster** | Not included: policy-dependent and severely regressed some output-heavy Airline cells because raw cancelling updates remain buffered until host drain. |
| Planner-proven redundant non-recursive IDB clamp removal | **0.5% faster** | Not included: proof and tests were sound, but too little real-workload benefit for the planner/API complexity. |
| Compact inline inspectors + flush only touched inputs | **0.1% slower** | Not included. |
| Disable string interning | **9.0% slower** | Rejected; keeping interning is beneficial. |

The proven-IDB prototype included the unsafe projection counterexample
`Src(x) :- Edge(x, y)` and removed clamps only under a conservative
single-producer/injective-rule proof. The output-netting prototype matched
baseline deltas across randomized recursive, negation, aggregation, nullary,
one-worker, and four-worker runs. Neither result justified expanding this PR.

This reinforces the current scope: the two options in this PR are the largest
broad, low-risk fixed-overhead wins found so far. The branch remains unchanged.

PR #321, opened 2026-09-01
feat: harden and optimize generated evaluators

nilspalumbo:
## Summary

- fail closed on missing, malformed, invalid-UTF-8, truncated, or extra-column input in batch and incremental generated evaluators
- preserve exact symbol bytes, including leading and trailing whitespace
- stop loading later file relations and inline facts after a fatal input error; the generated batch/epoch remains atomic and publishes no partial result
- make generated flow planning deterministic and remove redundant cross-stratum transformations and arrangements
- harden atomic generated-project installation, worker failure propagation, macOS signing, and release packaging
- retain the generated crate's compatible `flowlog-runtime = "0.3"` requirement; local and CI testing use the workspace `[patch.crates-io]` override until the next runtime release

The branch is rebased directly onto the current `main-next` integration branch.

## Validation

- `cargo fmt --all -- --check`
- `RUST_MIN_STACK=33554432 cargo test --workspace`
- generated binary fixture harness: all 120 fixtures pass
- focused final incremental `file_load` fixture: pass
- Sasty production evaluator build and missing/malformed-input smoke checks: pass
- packaged Sasty Flowlog findings and statistics exactly equal Soufflé across the Python default/context, Python review, and JavaScript review parity cases

The larger Rust stack is needed only for the existing parser deep-nesting test on this local macOS host. The commit includes a DCO sign-off.


HarukiMoriarty:
Hi Nils, thanks for the detailed PR and validation work! We use main-next as our integration branch, with changes flowing to main afterward. Could you please update your branch against the latest main-next, resolve the merge conflicts, and retarget this PR to main-next?

nilspalumbo:
Whoops, thanks for the catch! It's been rebased onto the right branch.

HarukiMoriarty:
Hi @nilspalumbo, thanks for working on this. I agree that FlowLog should fail loudly on input errors. Silently skipping malformed input and reporting success with incomplete results is not behavior we want.

Could we narrow this PR to input validation and error propagation? The planning/codegen changes, binary installation/macOS signing, and shell compatibility fixes would be easier to review separately.

For the input-handling changes, I’d like us to:

  - Consolidate repeated failure-handling logic where it shares the same behavior.
  - Define the failure boundary clearly: batch execution should fail without publishing partial results. For
    incremental execution, we should decide whether an input error rejects the current transaction or terminates the
    entire evaluator.
  - Add regression tests for malformed input after valid rows, failures across multiple input files, and multi-worker
    execution, checking for partial results and hangs.

One detail also needs clarification: the description says symbol whitespace is preserved, but the string parsing paths still call trim(). Changes to whitespace semantics would be easier to review separately as well.

If you’re happy with this direction, I can take care of narrowing the PR and making these adjustments on our side.

